PR #147, opened 2026-05-23
chore(eval): forward-looking judge API surface (stamp_judge_model + structured parser)

7xuanlu:
## Summary

Lays two **forward-looking API surfaces** for a follow-up PR that re-architects the batch judge pipeline. Neither is wired into a production path in this PR.

**Why this scope:**

Plan B (`docs/superpowers/plans/2026-05-22-eval-semantic-gaps.md` — gitignored) targeted three Plan A follow-up gaps: `judge_model` always `None`, `embedder_revision` placeholder `"768d"`, free-form judge text. Tasks 2 and 3 collided with codebase realities:

- **Task 2 dropped:** the FastEmbed cache → HF-snapshot-hash mapping is a one-day chore better done when same-dim embedder swaps become a real comparability problem.
- **Task 3's prompt-suffix change was reverted** (`5b1f618d`) after an adversarial review caught that `task_judge_prompt`'s new JSON-output instruction contradicted the base "answer yes or no" body, and the downstream parser at `judge.rs:1186` (`lower.starts_with("yes")`) would silently score 0 for every tuple. The type machinery survives as a future-API; the prompt body is unchanged.

**What's actually shipped:**

1. **`stamp_judge_model(env, model)` helper** in `eval/judge.rs` — populates `ReportEnv.judge_model` on `LocomoReport` / `LongMemEvalReport`. Today, the runners that own those reports don't call the batch judge directly (judge runs in `eval_harness.rs` integration tests, results land in a separate `JudgedE2EReport`), so this is a forward hook for whenever the runner + judge pipelines share a session.
2. **`JudgeVerdict` enum + `StructuredJudgeOutput` struct + `parse_judge_output`** — 3-tier parser (direct JSON → markdown-fenced JSON → legacy exact-string fallback). Not yet wired into `judge_with_batch_api`. The future wiring requires either (a) reconciling the "yes/no" vs "correct/incorrect/partial" verdict vocabularies, or (b) dispatching different parsers per prompt builder. Out of this PR's scope.

## Test Plan

- [x] `cargo test -p origin-core --lib eval::judge` — 7/7 (3 original + 4 edge-case regression tests for the legacy-string false-positive fix)
- [x] `cargo test -p origin-core --lib eval::locomo` — all green incl. new `locomo_env_records_judge_model_when_batch_runs` contract test
- [x] `cargo clippy -p origin-core --all-targets` — clean
- [x] Pre-commit hook (fmt + clippy on changed crates) passed
- [x] Adversarial review of integrated impl run + blocker reverted before push

## Notes for reviewers

- The 4 commits in this branch will squash on merge. Title prefix is `chore:` so release-please does NOT bump version — this PR adds public API surface but does not change any production behavior, so a version bump would mislead changelog readers.
- Follow-up PR scope (queued, not blocking this one): wire `judge_with_batch_api` to call `parse_judge_output`, add `Yes`/`No` verdict variants to `JudgeVerdict`, and stamp `judge_model` from inside the wired flow.

### DIFF
--- a/crates/origin-core/src/eval/judge.rs
+++ b/crates/origin-core/src/eval/judge.rs
@@ -1204,3 +1204,11 @@ pub async fn judge_with_batch_api(
     );
     Ok(results)
 }
+
+/// Stamp the judge model id on a Report's env (no-op when env is None).
+/// Used by runners to record which judge produced the answer-quality metrics.
+pub fn stamp_judge_model(env: &mut Option<crate::eval::report::ReportEnv>, model: &str) {
+    if let Some(e) = env.as_mut() {
+        e.judge_model = Some(model.to_string());
+    }
+}
--- a/crates/origin-core/src/eval/judge.rs
+++ b/crates/origin-core/src/eval/judge.rs
@@ -24,13 +24,7 @@ pub fn task_judge_prompt(
     ground_truth: &str,
     answer: &str,
 ) -> String {
-    let base = lme_anscheck_prompt(category, question, ground_truth, answer);
-    format!(
-        "{}\n\nRespond with a single JSON object on the final line:\n\
-         {{\"rubric_scores\": {{\"<criterion>\": 0.0_or_1.0, ...}}, \"verdict_reason\": \"<one sentence>\", \"verdict\": \"correct\" | \"incorrect\" | \"partial\"}}\n\n\
-         The rubric_scores keys must match the criteria listed above. Do not wrap the JSON in markdown fences. Do not add prose before or after the JSON object.",
-        base
-    )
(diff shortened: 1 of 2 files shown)


7xuanlu:
Closing — adversarial review surfaced that this PR is forward-only scaffolding (stamp_judge_model + parse_judge_output have zero production callers; the prompt-suffix change that would have wired them was reverted as a silent-zero regression). Wiring requires reconciling the LME yes/no verdict vocabulary with StructuredJudgeOutput's correct/incorrect/partial — that's a real design call, not a mechanical add. Branch stays on remote (`feature/eval-semantic-gaps`) for future revive. Will land the helper + parser in the same PR that does the wiring.

PR #180, opened 2026-05-25
perf(ci): cache + sccache + thin LTO + zigbuild for ~50% CI speedup

7xuanlu:
## Summary

Targeted at the 30-min Linux/Windows tail observed during the v0.7.0 release iteration. Expected total CI: 15-20 min (down from ~30).

Four levers, all community-proven for cross-platform Rust release matrices:

| Change | Where | Expected saving |
|---|---|---|
| 1. \`cache-all-crates: "true"\` on Swatinem rust-cache | ci.yml + release.yml | 2-3 min on cold-cache restore |
| 2. \`mozilla-actions/sccache-action\` + \`RUSTC_WRAPPER=sccache\` | ci.yml test + release.yml native + cross | 5-8 min on cold compile (C/C++ deps: ring, openssl-src, llama-cpp-sys) |
| 3. \`[profile.release]\` \`codegen-units=16\`, \`lto=\"thin\"\`, \`strip=\"debuginfo\"\` | Cargo.toml | 30-50% off link time |
| 4. \`cargo-zigbuild\` replaces \`cross\` for Linux targets | release.yml | 3-5 min off Linux jobs (no Docker container) |

Plus: zigbuild dodges the entire cross-0.2.5 Ubuntu-16.04-OpenSSL chain that consumed several PRs during v0.7.0.

## Tradeoffs

- Release binary 5-10% larger (codegen-units + thin LTO instead of fat).
- One new action dep (sccache-action) and one new tool (cargo-zigbuild + zig).
- Cross-platform cross-compile semantics with zigbuild are simpler but not identical to cross — needs validation that QEMU smoke step still works.

## Test plan

- [ ] CI on this PR runs to green (validates Swatinem cache + sccache wiring).
- [ ] Manual release.yml workflow_dispatch against v0.7.0 (or v0.7.1 if cut) to validate cargo-zigbuild + sccache + thin LTO release builds.

PR #181, opened 2026-05-25
fix(deps): use ort-download-binaries-native-tls (not rustls variant)

7xuanlu:
fastembed's `ort-download-binaries-rustls-tls` doesn't trigger ort-sys's prebuilt-onnxruntime download in cross. Switch back to native-tls variant; workspace openssl-sys vendored handles the system-OpenSSL fallout.


7xuanlu:
Combined with PR #180 into PR #182.

PR #183, opened 2026-05-25
perf(ci): skip Windows matrix when not Windows-affecting + fail-fast

7xuanlu:
Follow-up to PR #182. Three free-tier wins on the Windows tail (~31min after #182).

## Changes

1. **\`fail-fast: true\`** — cancel matrix on first failure. Real bugs failing Linux/macOS no longer waste 20-30min on Windows hitting the same failure.

2. **Drop workspace lib tests on Windows.** Linux + macOS catch the same cross-platform unit logic. Windows-specific behavior (paths, mcp_add, schtasks) is already covered by the \`Integration tests origin-cli + origin-server\` step (runs on all 3) and the schtasks install round-trip E2E. ~25min Windows saved per PR.

3. **Skip Windows job entirely** when no Windows-affecting files changed. New dorny \`windows\` paths-filter + \`detect-changes.test-oses\` JSON output. Most PRs (docs, eval, server internals) skip Windows entirely.

## Expected

- Typical PR: 31min → 14-15min (Linux floor)
- Windows-touching PR: 31min → 10-12min (Windows runs but skips workspace lib tests)
- Real-bug PR: 31min → 8-10min (fail-fast cancels Windows after Linux/Mac fail)

## Release pipeline

release.yml unchanged — release builds still cover all 5 targets on tag push.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,10 @@ jobs:
       rust:    ${{ steps.filter.outputs.rust }}
       plugin:  ${{ steps.filter.outputs.plugin }}
       npm:     ${{ steps.filter.outputs.npm }}
+      windows: ${{ steps.filter.outputs.windows }}
+      # Test matrix OSes as a JSON array. Windows is included only when
+      # Windows-affecting files changed (saves ~30min per PR otherwise).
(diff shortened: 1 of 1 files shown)


7xuanlu:
Combined into PR #186 — both fixes ship together as one CI follow-up.

PR #263, opened 2026-06-14
fix(eval): CE answer-accuracy 2-arm gate (event_date + power)

7xuanlu:
## What

Lands the **decision-critical** piece of the LongMemEval CE-default-on work: an honest 2-arm (CE-off vs CE-on) **answer-accuracy** harness, with the temporal substrate made live and the statistical test power-aware. This PR is what the gated CE decision run executes against.

Split out of an over-scoped plan after an adversarial 3-lab council (2026-06-13); the de-fork / provenance-migration / 3rd-graph-arm work is deliberately deferred to follow-up PRs (they are not on the CE *sign-of-delta* critical path).

## Commits

1. `feat(eval)` — **2-arm CE answer harness foundation**: per-question isolated scenario DBs (each Q sees only its own sessions = LME-correct), on-device Qwen3.5-9B answers at temp 0.0, paired McNemar over Haiku-judged answers via the `claude -p` CLI (no API spend). Judge guard relaxed to *warn* on `ANTHROPIC_API_KEY` (billing trap) instead of requiring it.
2. `fix(eval)` — **temporal gate + McNemar power** (this PR's hardening):
   - **T1.1** inject per-session `event_date` into each per-Q seed, then `assert_feature_substrate_live("temporal")` so a starved temporal substrate **refuses loud** instead of emitting a misleading null (eval ONE-contract). Reuses the canonical `event_date_map` + `set_event_dates_by_source_id` helpers — **no training-serving skew**; `source_id` keys are byte-identical to the seed closure.
   - **T1.2** `McNemarReport` gains `n_discordant` (b+c) + `underpowered` (b+c<10), printed by the judge so a null delta at small n (first gated run = 50 Q) is not misread as "CE doesn't help".

## Honest scope note (caught in review)

The `event_date` injection is **substrate-prep + a liveness gate, not a direct answer fix**: `event_date` is not rendered into the CE A/B context and no temporal-ranking flag is on this path, so dates influence neither ranking nor the prompt **today**. Temporal-reasoning answers therefore stay date-blind — but the **CE delta remains valid** because both arms see identical content-only context (CE rerank is the only difference). Rendering dates / wiring a temporal lever is future work, not this PR.

## Council flaws addressed

- **Flaw 3** (temporal starvation) → T1.1 inject + gate.
- **Flaw 4** (power / underpower at n=50, single-pair) → T1.2 (Holm/FWER across pairs is deferred to the 3-arm PR).

## Tests

- `event_date_injection_lands_on_seeded_rows` — deterministic contract test guarding the `source_id`↔`event_date_map` key-alignment silent-zero (real DB, real upsert, asserts exactly 1 row updated).
- `mcnemar_underpowered_small_discordant` / `_powered_sufficient_discordant` / `_boundary_at_ten` — pin the `<10` floor incl. the exact boundary (off-by-one guard).
- `ce_ab_seed_injects_event_date` — `#[ignore]` GPU runtime proof (Metal + minutes; runs in the gated run, not CI).

Full verify green: `cargo fmt --check`, `cargo clippy -p origin-core --all-targets --features eval-harness -- -D warnings`, `cargo test -p origin-core --lib eval` = **525 passed / 0 failed**.

## Not in this PR (gated)

The headline CE decision number itself. Running this harness (first-50 oracle N=1 → resume to 500) is **held for explicit trigger**; single-run = scaffold, headline claims need N≥3 + stddev per eval-citation discipline.

### DIFF
--- a/crates/origin-core/src/eval/answer_quality.rs
+++ b/crates/origin-core/src/eval/answer_quality.rs
@@ -580,6 +580,7 @@ pub async fn run_e2e_locomo_eval(
                         answer,
                         context_tokens: origin_ctx_tokens,
                         category: String::new(),
+                        question_id: String::new(),
                     });
                 }
                 Err(e) => {
@@ -634,6 +635,7 @@ pub async fn run_e2e_locomo_eval(
                             answer,
                             context_tokens: compressed_ctx_tokens,
                             category: String::new(),
+                            question_id: String::new(),
                         });
                     }
                     Err(e) => {
@@ -673,6 +675,7 @@ pub async fn run_e2e_locomo_eval(
                             answer,
                             context_tokens: replay_ctx_tokens,
                             category: String::new(),
+                            question_id: String::new(),
                         });
                     }
                     Err(e) => {
@@ -712,6 +715,7 @@ pub async fn run_e2e_locomo_eval(
                         answer,
                         context_tokens: 0,
                         category: String::new(),
+                        question_id: String::new(),
                     });
                 }
                 Err(e) => {
@@ -817,6 +821,7 @@ async fn generate_e2e_answers_for_question(
             answer,
             context_tokens: flat_tokens,
             category: category.to_string(),
+            question_id: String::new(),
         });
     }
 
@@ -861,6 +866,7 @@ async fn generate_e2e_answers_for_question(
             answer,
             context_tokens: structured_tokens,
             category: category.to_string(),
+            question_id: String::new(),
         });
     }
 
@@ -2189,6 +2195,7 @@ pub async fn run_fullpipeline_locomo_batch(
                 answer: answer.clone(),
                 context_tokens: meta.context_tokens,
                 category: meta.category.clone(),
+                question_id: String::new(),
             });
             matched += 1;
             if matched.is_multiple_of(10) {
@@ -2684,6 +2691,7 @@ pub async fn run_fullpipeline_lme_batch(
                 answer: answer.clone(),
                 context_tokens: meta.context_tokens,
                 category: meta.category.clone(),
+                question_id: String::new(),
             });
             matched += 1;
             if matched.is_multiple_of(10) {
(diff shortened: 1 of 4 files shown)


7xuanlu:
## Adversarial review pass (post-open)

Ran a 3-lab council (Claude + Codex + Gemini) on the *implementation* (not just the plan), then an independent Opus review on the resulting diff. Findings + fixes, all landed on this branch:

**Council (2-member; the codex member stalled on a tool-availability issue, diagnosed separately):**
- **F1 (real bug)** — `paired_answer_mcnemar` paired the OFF/ON arms by keying on question **text**, so two distinct questions sharing text collapsed via `HashMap` overwrite, silently corrupting b/c/p. Fixed: thread a `question_id` (serde-default for back-compat) through `JudgmentTuple` → judge → `JudgmentResult` and key the pairing on it (fallback to text when empty). Regression test `mcnemar_pairs_on_question_id_not_text`. Commit `1c759dd4`.
- **F2 (confound)** — the entry guard neutralized only `ORIGIN_GRAPH_MEMORY_STREAM`. Extended to refuse on any arm-asymmetry confounder: graph stream + `ORIGIN_RERANK_SKIP_PREFERENCE` + `ORIGIN_ENABLE_TEMPORAL_SOFT_BOOST` + `ORIGIN_ENABLE_TEMPORAL_FILTER`, naming the offenders. Commit `1c759dd4`.
- **F3 (robustness)** — the per-question `event_date` injection used `?`-propagation, so one source_id mismatch aborted the whole multi-hour run. Now logs + `continue` (skips that question) on a 0-row update. Commit `1c759dd4`.
- **F4 (framing)** — documented that this A/B isolates the cross-encoder (stream forced off both arms) and does **not** by itself prove the production default-flip (which couples CE-on with stream-off). Commit `1c759dd4`.

**Independent Opus review of the fix diff** found F1 was incomplete: the resume dedup (`done_questions`) and the judge-cache dedup still keyed on question text. Completed in `78b98ebd` (both now key on `question_id` with text fallback). Verified empirically that both fixtures (`longmemeval_oracle.json`, `longmemeval_s.json`) have **0 duplicate question texts**, so this was a dormant landmine — fixed correct-by-construction regardless.

Full gate green throughout: `cargo fmt --check`, `cargo clippy --all-targets --features eval-harness -- -D warnings`, `cargo test -p origin-core --lib eval` = 526 passed / 0 failed.

The headline CE decision run (first-50 oracle → 500) remains gated; single-run = scaffold, citation needs N≥3 + stddev.

PR #470, opened 2026-08-03
feat(m6): freeze the S0-97 relevance benchmark corpus

7xuanlu:
**Stacked on #469** (`m6-pr-b1`). Review that first; this PR's own diff is the last commit only.

PR-C slice C0: the frozen benchmark corpus that C1's relevance gates will measure against. No M6 runtime code is touched, so this reviews as *"is this corpus reproducible"* rather than *"is this sweep correct"*.

## Why this lands before C1

Artifact 6 §10's count limits are hard pass/fail and S0-99 forbids answering a benchmark failure by tuning a constant. Landing C1's thresholds against an ad-hoc fixture and retrofitting the corpus afterwards would make the first frozen run also the first run anyone could argue with. A gate that arrives after the code it gates has never once failed.

Artifact 6 states it in one line: *"A benchmark whose corpus is not reproducible is not a gate."*

## What ships

- `eval/m6_bench_corpus.rs` — the eleven `R-CORPUS-*` constants of S0-97 as named consts, seed `0x6D36_0000`, encoding `wenlan-m6-bench-corpus-v1`, a manifest digest over corpus + composition, refuse-on-mismatch.
- `tests/m6_relevance_bench.rs` + `tests/fixtures/m6_bench_corpus.sha256` (`aeee4a5a…b701a3da`).
- `pub(crate)` lift of M5's freeze machinery (`SplitMix64`, `sha256_hex`, `validate_sha256_hex`, `DigestWriter`). Mechanical, no behavior change, and the only edit made to M5's module.
- A separate `m6-platform` CI filter, its step, and its `drift_guard` tooth.

## Two decisions worth review attention

**Integer-only Zipf weights.** The degree distribution is Zipf(11/10) truncated at 5000, computed without `f64::powf`: `k^1.1 == k * k^0.1`, and `k^0.1` is an integer 10th root by binary search in Q20 fixed point. A frozen corpus must reproduce byte-for-byte on macOS and Windows — both scheduled by the `m6-platform` filter — and `powf` is not guaranteed bit-identical across platform libm. A last-ULP difference would surface as a hard digest mismatch on one platform only: the refusal would be correct, the cause near-unreadable.

**A separate filter, not a fold into `m5-platform`.** The M5 step's `run` is pinned verbatim to `--test m5_bench` by an existing tooth, so one filter cannot drive two bench targets without either unpinning that run — which is the tooth's entire content — or running the wrong bench on every M6 change. The fold is structurally unavailable, not merely untidy. The new tooth is purely additive: `m5_platform_inputs` keeps its exact seven paths and its `!=` comparison.

The fixture and the tooth are in one commit because `drift_guard` is fail-closed; a routed path landing a commit ahead of its expected set breaks the build for whoever pulls in between.

## Verification

- `cargo test -p wenlan-core --lib` — **4028 passed, 0 failed**, 33 ignored, 645.77s. Hash-pinned: sha256 of all seven touched files snapshotted before the run, `shasum -c` after, all seven OK, so the result is provably tied to the committed bytes.
- `cargo test -p wenlan-core --test m6_relevance_bench` — 5 passed
- `drift_guard` — 154 passed
- clippy `-D warnings` and `fmt --check` — clean

### RED controls

| Defect restored | What fails |
|---|---|
| M6 paths folded into the `m5-platform` filter | **both** teeth: `m5-platform focused-owner routing is not exact` and the M6 one — proof the new tooth is additive and M5's was not weakened |
| M6 step's `run` unpinned to `--test m5_bench` | `the focused M6 platform owners do not compile and run the frozen-corpus target` |
| `fixed_tenth_root` perturbed by one Q20 unit | the bench refuses on digest mismatch |

The third control is the one worth reading. Under that perturbation — the magnitude of a cross-platform libm difference — `the_corpus_regenerates_byte_for_byte` **still passes**, because it only compares two runs on the same machine. Only the checked-in fixture catches a systematic difference. A self-consistency check is blind to exactly the failure the freeze exists to prevent.

## One defect this PR's own gate caught

The fanout mean was drawn as an independent uniform over `1..=15`, which has mean 8 *in expectation* but realized 95,464 edges against a specified 96,000. Fixed by construction — a balanced, shuffled assignment where every value in the span appears equally often — rather than by relaxing the assertion, so `R-CORPUS-FANOUT` stays exactly checkable and matches how `R-CORPUS-RETFRAC` is handled.

https://claude.ai/code/session_01JmrDX8z66BwPFbL525xw91


7xuanlu:
Closing: the S0-97 frozen-corpus benchmark lane is not being pursued; the genesis-shadow flag's default-ON prerequisite list should drop this item accordingly.

PR #471, opened 2026-08-04
feat(m6): maintain a bounded co-citation relevance substrate

7xuanlu:
Third rung of the M6 PR-C ladder (`docs/plans/2026-08-03-m6-pr-c-maintenance-shadow-spec.md`). Stacked on #470; review that first.

## What this adds

The C1 relevance substrate: a decay-referenced co-citation estimator over independence groups, maintained by a **bounded** sweep plus an exact incremental applier.

- `m6_pair_stats` stores only `n11` + `distinct_group_count`. `n10`/`n01`/`n00` are derived against marginals in `m6_page_mass` / `m6_space_mass`, so one root activation rewrites a bounded slice instead of the whole table (`R-HUB-MAXPAIRS`).
- One deliberately unbounded read — the full re-reference pass — establishes the pin every bounded turn measures against. Q1 sanctions it; it carries no `QueryBudget` on purpose.
- `hub_weight(d) = min(1, 64/d)` over pages a group touches, cut at 64 identically in every path.

## Verification

- **S0-91 byte-equality oracle**: the whole pair table digested after each of eight mutations plus a randomized interleaving, incremental vs full recomputation, with a negative control.
- **S0-95 budget instruments**: decoded-row counter, query counter, `SQLITE_STMTSTATUS_FULLSCAN_STEP == 0`, `EXPLAIN QUERY PLAN`.
- `cargo test -p wenlan-core --lib` — **4072 passed, 0 failed**. Clippy clean under `--all-features`.

## Two defects found in review and fixed here

**Migration gap (P0).** `ensure_relevance_tables` is reached only from migration 109, behind `if version < 109`. Every shipped database is already at 110, so it would never see the two new marginal tables and the first relevance write would fail with `no such table: m6_space_mass`. All 4070 tests passed with the bug present because every fixture builds a fresh database or calls the DDL directly. Fixed with migration 111 plus the upgrade-from-110 fixture whose absence let it ship green.

**Pairless-page selection (correctness).** `candidate_endpoints` selected a page only when some `m6_pair_stats` row naming it was behind the generation. `apply_group_mutation` owns a group entering or leaving the eligible universe, so a group that stays eligible while gaining a page never reaches it — and a group supporting exactly one page emits no pair, so when it gains a second, neither endpoint has a row to be stale and the pair is invisible forever. That is a row a full recomputation emits and the incremental table never contains. Selection now also admits a co-supported page with no row at all, counting only pages inside the top-64 cut so a page ranked past it cannot be re-selected forever.

`groups_touching` now derives touched groups straight off the endpoint pages instead of building every group in the space and joining. Same result set, same statement count.

## Known gap, not fixed here

The C0 benchmark corpus cannot exercise the hub cap: `M6_HUB_DEGREES` sets *root* degree, while `hub_weight` keys on *pages touched*, which comes from a fanout drawn `1..=15`. Maximum pages-touched is 15, so `hub_weight` returns exactly `1.0` for all 12,000 groups and the `64/d` branch never executes — `R-HUB-TOPK` / `R-HUB-WEIGHT` / `R-HUB-MAXPAIRS` are untested by the benchmark. That is C0's corpus, so it is fixed on #470 rather than here.

https://claude.ai/code/session_01JmrDX8z66BwPFbL525xw91


7xuanlu:
Closing alongside #470: the bounded co-citation relevance substrate served the same benchmark lane, which is not being pursued.

PR #534, opened 2026-08-16
fix: make the repository buildable, testable, and pushable on Windows

7xuanlu:
## Problem

This repository could not be developed on Windows, and - more seriously - **could not be pushed from Windows at all**. Twelve independent defects, none of which any CI job can see: `app-check` runs on `macos-14` only, `app-windows-bundle` is `workflow_dispatch`-only, and the Rust test lanes are Linux and macOS. So `pnpm install`, `pnpm test`, `cargo clippy --all-targets`, and `cargo test --workspace --lib` have never executed on a Windows runner.

Found during a physical verification pass on a Windows 11 machine (x86_64-pc-windows-msvc).

## 1. Nobody on Windows could push (the blocking one)

`cargo test --workspace --lib` on Windows: **3989 passed, 116 failed**.

115 of the 116 have a single cause. Every repair entry point - `prepare_memory_reclassification`, `apply_repair`, `record_repair_verification` - calls `ensure_repair_artifacts_supported`, which returns `Validation("lint_repair_unsupported_platform")` on non-unix. That refusal is deliberate: the artifact store's privacy guarantee is POSIX mode bits and there is no Windows equivalent wired up. But the tests that drive a real repair were never gated to match, so they fail on a platform where the product is behaving exactly as designed.

That is not cosmetic. `.githooks/pre-push` runs the affected-package closure, and `ci_test_plan.py` is fail-closed: a workflow file, an unowned path, or any shared build input widens the plan to the full workspace. In practice essentially every push reaches this failure, so a Windows developer cannot push anything - including a one-line docs fix.

Fix: `#[cfg_attr(not(unix), ignore = "repair artifacts are unix-only")]` on those 115 cases, plus the invariant recorded next to the gate itself so the two stay in step.

`ignore` rather than `cfg` on purpose - the cases still compile everywhere, and a Windows run reports them as ignored rather than making them silently disappear. **macOS and Linux still run all 115.** If the platform gate is ever opened, the attributes come out with it.

Result on Windows: `119 passed; 0 failed; 115 ignored` for the repair subset, and the full workspace lib suite goes green.

Gating them gave five test files a `unix` cfg expression they did not carry before, and the CI routing contract is fail-closed about exactly that: a source with a platform cfg must be owned by both the `macos` and the `windows` path filter, so that a change to it schedules the lanes which can prove the gate on each side. `repair.rs`, `repair/**`, and `repair_plan/**` were already routed and already carried the same cfg; `repair_plan_tests.rs` and the four `db/repair_*_test.rs` files simply sit outside those globs. They are now listed next to them. `drift_guard` is the authority and it passes: `cargo test -p wenlan-core --lib drift_guard::`, 157 passed, 0 failed.

## 2. `cargo clippy --all-targets` could not pass on Windows

```
error: function `bind_addr_from_stub_host` is never used
  --> crates\wenlan-cli\tests\cli_integration.rs:343:4
  = note: `-D dead-code` implied by `-D warnings`
```

All eight callers are launchd/unix background-lifecycle tests gated on `#[cfg(target_os = "macos")]` or `#[cfg(unix)]`, so on Windows the helper is dead code and `-D warnings` fails the whole target. The lint job runs on Linux, so nothing caught it. Gated the helper the same way its callers are gated.

## 3. `pnpm install` failed outright

```json
"postinstall": "git config core.hooksPath .githooks 2>/dev/null && chmod +x .githooks/* 2>/dev/null || true"
```

pnpm runs lifecycle scripts through `cmd.exe` on Windows. That line is a parse error there, so `postinstall` failed and took the whole install with it.

Replaced with `node scripts/install-git-hooks.mjs`, which does the same two things and stays best-effort exactly like the `|| true` it replaces. Node is available by definition inside a lifecycle script. The `chmod` is skipped on Windows, where there is no POSIX execute bit and git ignores the mode anyway.

## 4. The `block-no-verify` agent hook failed closed on every Bash call

`.claude/hooks/block-no-verify.sh` hardcoded `/usr/bin/python3`. Git Bash on Windows has no such path, so the hook could not extract the command, hit its fail-closed branch, and blocked every Bash tool call in this repository - including ones with nothing to do with git.

Now probes `python3` then `python` from PATH, the same way `.githooks/pre-push` already does. Fail-closed behaviour is unchanged and still correct: no interpreter at all is still a block. Verified with the existing 8-case canary in `.claude/hooks/test-hooks.sh`, both the block and the allow directions.

## 5. `pnpm test` failed 16 cases, all POSIX-only dev-script tests

Every failure was in `scripts/dev-runtime.test.ts` or `scripts/prepare-sidecars.test.ts`, and every one shells out to the POSIX dev scripts: spawning `bash`, symlinking `/bin/sleep`, `chmod`ing a fake `lsof`, joining `PATH` with `':'`.

On Windows `bash` is not even guaranteed to be Git Bash. On a machine with WSL installed it resolves to `C:\Windows\System32\bash.exe` - the Linux distro - whose PATH carries no Windows `node` or `rustc`, so the scripts die with `node: command not found`.

Those cases are now gated on `process.platform`. The package.json, Tauri-config, and script-text contracts in both files stay platform-neutral and keep running everywhere. macOS and Linux keep all 28 cases.

Three of the gated cases (the `it.each` production-identity guards) did pass on Windows, but only by accident - WSL bash got far enough to hit the refusal branch, and one of them flaked under load. Passing through the wrong interpreter is not coverage, so they are gated with the rest.

## 6. Two genuinely wrong tests

- `eval::report::tests::encode_baseline_path_layout` built its expectation as a literal `/`-joined string while the function under test uses `Path::join`. It asserted the separator, not the layout. Now builds the expectation with `join` too.
- `SpacesOverview.management.test.tsx` clicked the row's actions trigger as soon as the mutation had been *called*, while the row was still the disabled pending editor - so the click landed on a disabled button and the following query threw. Reproduced at roughly 2 failures in 3 runs. Now waits for the trigger to be enabled first. Verified with 8 consecutive green runs.

## 7. Pre-push ran the full workspace lib suite for changes it could not be informative about

Fixing 1 through 6 made the suite pass on Windows, but it left the underlying gate wrong. The planner is fail-closed, so a workflow file, an unowned path, or any shared build input widens the plan to the whole workspace. At push level that meant `cargo test --workspace --lib` - over an hour on this machine - for exactly the changes least likely to break a specific package.

On macOS and Linux it bought nothing: `workspace-lib-required` runs the same suite on every PR and main push, so the local run was duplicate work rather than extra coverage. On Windows it did buy something - the required Windows job skips the workspace lib suite, so a widened local plan was the only lane that ever executed a `#[cfg(windows)]` unit test. That coverage is given up here knowingly; section 11 and `docs/test-layers.md` both record it. What the gate cost was worse than what it bought: it is what turned a platform-specific test failure into "this platform cannot push at all".

A widened plan now prints `local checks: plan widened to the full workspace; CI owns that suite` and stops at Clippy, which stays fail-closed. Narrow plans are untouched: an affected-package change still runs its own lib tests, a directly edited integration target still runs alone, and isolated unit-test owners still run by exact filter. `docs/test-layers.md` records the reasoning next to the existing note on why pre-push does not run coverage.

`scripts/ci_test_plan.test.py`: 93 tests, OK.

## 8. Every dev script ran under WSL, so the app could not be started at all

`package.json` spelled the dev scripts as `bash scripts/<name>.sh`. On Windows that resolves through PATH, and on any machine with WSL installed the first `bash` on PATH is `C:\Windows\System32\bash.exe` - the Linux distro, not Git Bash. The scripts then ran inside Ubuntu, where the Windows toolchain does not exist:

```
> bash scripts/prepare-sidecars.sh
scripts/prepare-sidecars.sh: line 30: rustc: command not found
```

`pnpm tauri dev` and `pnpm tauri build` both reach this through `beforeDevCommand`/`beforeBuildCommand`, so **the desktop app could not be started from source on Windows at all**.

`scripts/run-bash.mjs` resolves Git Bash explicitly on win32 - from `WENLAN_BASH`, the standard install roots, or the `bin` directory beside whichever `git` is on PATH - and is a plain passthrough to `bash` everywhere else. Seven scripts rewired: `dev:daemon`, `dev:all`, `prepare:sidecars`, `prepare:sidecars:tauri-build`, `release:daemon`, `update-all`, `clean:dev`. macOS and Linux behaviour is unchanged.

The two contract tests that pinned the old command strings now pin the new ones, and the prepare-sidecars fixture root copies `run-bash.mjs` alongside the shell scripts it launches.

## 9. `pnpm dev:all` could not start the daemon on Windows

With the interpreter fixed, the dev runtime itself was still POSIX-only, and it failed the worst way a script can: `ELIFECYCLE Command failed` with no message at all.

`lsof` does not exist in Git Bash. Under `set -e` the failing command substitution in `listener_pid="$(listener_pid_for_port ...)"` exited the script with 127 and printed nothing, because the only diagnostics on that path were already redirected to `/dev/null`. Behind that were three more:

- MSYS `ps` has neither `-p` nor `-o`, so the ownership check could not read a process image path at all.
- A job backgrounded from bash carries **two identities**. `$!` yields the MSYS pid; every Windows listener and process table reports the Windows one. The listener-ownership check compared one against the other and could never match.
- Windows cannot deliver `SIGTERM` to a console process that owns no console, so `kill` could not stop the daemon the script had just started.

Each primitive keeps its original POSIX implementation and gains a Git Bash branch - `netstat` for the listener table, `tasklist` for liveness, `ps -W` for the image path, `taskkill` for termination. The runtime resolves and records the **Windows** pid, so the pid file, the listener owner, and the identity check all compare like with like.

Resolving that pid needs care, and the first attempt got it wrong in a way only a real run exposed. `nohup env ... server &` is a chain of MSYS processes ending in an `exec` of a native binary, and MSYS spawns a fresh Windows process for that last step. Until it does, the MSYS pid still maps to the WINPID of `env`, which then exits - so reading the first WINPID that appears records a pid that is dead moments later:

```
wenlan-server.pid        8968          <- recorded
netstat -ano :17895      ... 8872      <- actually listening
error: refusing to stop PID 8968 because it is not .../daemon/wenlan-server.exe
```

The daemon started fine, was disowned by its own launcher, and leaked. The lookup now waits for a single `ps -W` row to name both the pid and the program that was launched, and the failure branch kills the MSYS job rather than leaving an unidentifiable daemon on the port.

The safety guard is unchanged in substance and was re-proved on Windows: with a live foreign pid recorded, `stop` refuses it and leaves it running.

```
error: refusing to stop PID 15088 because it is not C:/.../daemon/wenlan-server.exe
"PING.EXE","15088","Console","1","5,488 K"
```

Three further Windows-only defects on the same path:

**`print_config` emitted MSYS paths.** `WENLAN_DATA_DIR=/tmp/wenlan-app-dev/<id>/data` is handed to a Windows binary that calls `std::fs::canonicalize` on it in `validate_debug_runtime_isolation`, which cannot resolve it. The three directory values now convert at that boundary only; the script keeps working in the shell's own form.

**Sidecar staging deadlocked against the daemon.** Tauri's build script copies the staged sidecars and their runtime DLLs into the cargo target directory. A daemon running out of that directory holds `onnxruntime.dll` open and the copy fails with `os error 32`, so `dev:all` fought its own daemon. The Windows path now runs the daemon from a private copy under the dev state directory, with the DLLs the loader resolves from the executable's own directory copied alongside it.

**The server path ignored `CARGO_TARGET_DIR`**, which is how a Windows checkout stays under `MAX_PATH`. It now uses the same `${CARGO_TARGET_DIR:-$backend/target}` resolution `prepare-sidecars.sh` already uses. Without it the staged daemon was silently a *stale* build - worse than a clean failure.

`refuse_production_path` also gains the two Windows roots the app already enforces in `production_runtime_roots`, appended only when `LOCALAPPDATA` is set so an empty value cannot canonicalize to the working directory and refuse the whole checkout.

macOS and Linux take the original branch of every switch. The only POSIX-visible change is the target-root resolution, which is identical when `CARGO_TARGET_DIR` is unset.

Windows receipts:

| check | result |
|---|---|
| `pnpm dev:daemon` | starts and reports a healthy daemon on the isolated port |
| re-run `start` | reuses it; `start-for-session` exits 10 |
| `pnpm clean:dev` | stops it, frees the port, clears the recorded state |
| `stop` with a foreign live pid | refused, decoy left alive |
| `pnpm dev:all` | runs end to end and opens the app |
| `vitest dev-runtime + prepare-sidecars` | 10 passed, 28 skipped, 0 failed |

The 28 skipped cases are the POSIX-only ones gated in section 5; macOS CI still runs all 38.

## 10. The three remaining Windows gaps

An earlier revision of this description carried these as "found on the verification pass, not fixed here". They are fixed now, so that list is gone rather than shortened.

**macOS-only copy shown on every platform.** Five user-facing strings asserted the user was on a Mac: the intelligence step's "Local models keep inference on your Mac", the agents step heading "Found on this Mac" and its empty state "No AI tools found on this Mac.", and both Remote Access tunnel notes ("changes when your Mac sleeps or restarts"). Wenlan ships for macOS, Linux, and Windows, so all five were wrong on two of the three.

They now say "machine" - already the word the neighbouring copy in the same wizard step uses (*"Nothing else leaves your machine"*, *"inference stays on your machine"*). zh-CN and zh-TW follow the term each locale already used for the same idea, 设备 and 裝置. The `setup.connect.detectedOnMac` key is renamed `detectedOnMachine` so the key does not outlive the copy; both call sites and all three locales move together.

**`cloudflared` failed late and undocumented.** It is a hard requirement of `prepare-sidecars.sh` - Wenlan ships it as a Tauri `externalBin` for Remote Access and there is no fallback - but the check sat *after* the cargo build, at the install step. A checkout without it built the entire backend first and only then failed, and nothing in `docs/` or `README.md` said to install it. The check now runs where the source is resolved, before any build, and the message carries the download URL. `--print-paths` is unaffected: it still reports the placeholder rather than failing, which is what the Tauri `beforeBuildCommand` probe depends on. `CONTRIBUTING.md`'s Requirements block now names `cloudflared` and pnpm as desktop-app prerequisites, and says the daemon, CLI, and MCP crates need neither.

**The dev MCP socket was never cleaned on Windows.** `dev-all.sh` guards its cleanup with `[[ -S ... ]]`, which is what stops a mis-set `WENLAN_DEV_TAURI_MCP_SOCKET` from turning the call into an arbitrary `rm`. Git Bash has no socket file type, so that test is never true on Windows and anything left at that path would survive every run. The guard now falls back to a regular-file test **on Windows only**. The path lives inside the dev state directory, which `dev-runtime.sh` already refuses to aim at a production root, so removing a plain file there is as bounded as removing the socket is elsewhere.

Receipts: `pnpm test` 175 files, 1898 passed, 29 skipped, 0 failed. `prepare-sidecars.sh` with `cloudflared` off `PATH` and `CLOUDFLARED_BIN` unset fails immediately with the three-line message and never reaches cargo; `--print-paths` still exits 0. The socket cleanup was exercised both ways on Windows: an absent path is a no-op returning 0 (it runs under `set -e`), a plain file at the socket path is removed.

## 11. What independent review caught

Independent closure review ran against the whole branch, seven rounds of it. Everything it found is fixed, and the later rounds confirmed the earlier rounds' fixes against Microsoft's own process-object contract and against `origin/main`.

### Round 1

**A POSIX hook was committed with CRLF.** The Windows fix to `.claude/hooks/block-no-verify.sh` landed with `\r\n` on every line; `main` had LF. Nothing on Windows noticed. On macOS and Linux the configured `bash "$R/$H"` invocation reads the carriage returns as part of each token - `set -euo pipefail\r`, `then\r` - and the hook is fail-closed, so the result is that every Bash tool call in this repository is blocked on those platforms. Exactly the failure this PR removes, moved to the other two.

The cause is local and would recur: this checkout carries `core.autocrlf=false` over a working tree written CRLF, and git's stat cache hides the mismatch until something forces a re-read, so any POSIX script edited here is one `git add` away from the same commit. The file is back to LF and `.gitattributes` now makes it structural - `*.sh`, `*.bash`, `.githooks/*`, `.claude/hooks/*` are `text eol=lf`. Those are the files where a carriage return is fatal rather than cosmetic and where the failure mode is fail-closed. No other tracked file is affected and every `*.sh` is already LF in the index, so the attributes add no renormalisation.

**The Windows kill was pid-only.** `stop_runtime` proves ownership - image path and listening port both match the recorded server - and then called `taskkill /F /PID n`. Two separate calls, so Windows is free to retire the daemon and reassign its pid in between. The startup fallback was the worse half: when `windows_pid_for_job` gave up it sent SIGKILL to the MSYS pid from `$!`, after a ten-second window, to a pid MSYS may already have recycled, with no revalidation.

The first fix added `taskkill`'s image-name filter. Round 2 rejected that as insufficient and it was replaced; see below.

**Windows-gated unit tests have no required-CI execution.** True. `workspace-lib-required` runs on Linux and macOS only, by design - the Windows job skips the workspace lib suite because Linux and macOS cover the same cross-platform unit logic in a third of the wall-clock, so a `#[cfg(windows)]` unit test has no required-CI execution anywhere. The first response called this purely pre-existing. Round 2 showed that was wrong; see below.

### Round 2

**An image *name* is not an identity.** `taskkill` filters on a pid and on an image name, and a production install and every other worktree's dev daemon are all called `wenlan-server.exe`. The reuse window was narrowed, not closed - and checking the full path in a separate call would not have closed it either.

So the kill now opens the process before it judges it. `powershell.exe` resolves the pid to a `Process`, forces its `Handle` to materialize, and only then compares `MainModule.FileName` to the recorded server path and calls `Kill()`. Windows keeps a pid reserved for as long as any handle to it is open, so the path that is read and the process that is killed are provably the same one rather than merely close together. An identity that cannot be read kills nothing: the failure mode is a daemon that does not exit, which the caller already reports, not a stranger that disappears. Neither helper kills a process tree now, which matches POSIX `kill -KILL` and is exactly what should not follow an identity check that covers only the root.

Proven on this machine: a live `PING.EXE` survives a kill aimed at its pid with a non-matching path, survives one naming a sibling in the same directory, and dies to its own full path; a dead pid is a no-op.

**The startup fallback could leak an unrecorded daemon.** It took a single snapshot of the dev port. A daemon that had not bound yet was left running with nothing recording it - `dev-all.sh` only registers the runtime once `start` succeeds - so no later `stop` could find it, and it went on to take the port unmanaged. It now sweeps by image identity, which needs neither the pid nor the port to have settled, for as long as the health probe downstream would have waited, and says so explicitly when it still cannot.

**The coverage claim was wrong.** Section 7's note said pre-push never covered those Windows-gated tests. It did. Before the widened plan was trimmed it scheduled `cargo test --workspace --exclude wenlan-app --lib` and the `wenlan-server` binary tests, so a Windows developer pushing a broad change executed every `#[cfg(windows)]` unit test on the only platform that compiles them. Trimming that plan is what left them with no lane at all. `docs/test-layers.md` now says that, names what still covers Windows at a coarser grain, and prices closing the gap - so the trade is a decision on the page rather than a claim that nothing was lost.

**The new Node launchers had no app-check route.** `scripts/**` already schedules Rust CI, but app-check is the only job that installs the package. It is therefore the only job that runs the postinstall hook installer, and the only one that runs the Vitest suites asserting on `dev-runtime.sh`, `dev-all.sh`, `run-bash.mjs` and `install-git-hooks.mjs`. A single-file change breaking Git Bash discovery or hook installation scheduled Rust CI and reported green. All four are now in the `ci.yml` `app` filter and in `APP_JOB_FILES`, which `ci_test_plan.test.py` cross-checks against each other.

### Rounds 3 and 4: one directory, many spellings

Round 3 re-derived the round 2 fixes from primary sources - Microsoft's process-object lifetime contract and .NET's `Process.Handle`/`Kill()` implementation for the pinned kill, `origin/main` itself for the corrected pre-push history - and found them sound. Both rounds then found the same class of defect in code this PR adds.

**The production-root guard compared spellings, not directories.** It canonicalizes a candidate path and each production root and asks whether one is inside the other, as strings. On Windows that is not enough, and every spelling it could not reduce was a way past it:

- **Case.** `realpath` hands back whatever casing the caller wrote, so `%LOCALAPPDATA%\WENLAN` and `%LOCALAPPDATA%\wenlan` arrived as two strings naming one directory. (Round 3.)
- **DOS 8.3 aliases.** `C:\Users\LONGUS~1\AppData\Local\wenlan` and its long form, likewise. (Round 4.)
- **The `\\?\` extended-length prefix**, and **trailing dots or spaces** - Win32 drops those from every component, so `...\wenlan.` opens `...\wenlan`. Found by sweeping the rest of the class rather than waiting for a fifth round.

Each one meant `WENLAN_DEV_DATA_DIR=<that spelling>` started the dev daemon against the production data directory. The desktop app's own guard is too late by then - the daemon is already up - and `dev:daemon` never reaches it at all.

Reducing these by hand is a losing game; the first two fixes each closed one spelling and left the next. So `canonicalize_path` now asks the OS instead: `fs.realpathSync.native` is `GetFinalPathNameByHandle` on Windows and answers with the real on-disk path, settling case, aliases, the extended-length prefix and reparse points in one call, and is `realpath(3)` on POSIX, which is what this always used. Trailing dots and spaces are dropped before the call, because Node resolves through the `\\?\` form where they are literal rather than stripped. `path_is_within` still folds case on Windows as belt-and-braces, since a production root that does not exist on a given machine is not resolved by the OS call either.

### Rounds 5, 6 and 7: neither rewriting nor pretending

**The trailing dot/space strip rewrote paths where those characters are real.** `\\?\` and `\\.\` mean *do not normalize this*; inside them a trailing dot is part of the name, not something Win32 discards. The strip did not know that, so `\\?\C:\scratch\dev.\data` was reduced to `C:\scratch\dev\data` - a different directory, which the dev daemon would then have written into if it existed. (Round 5.)

Exempting verbatim paths from the strip looked like the fix and was not. `realpathSync.native` answers *without* the prefix, so the exemption preserved the literal components into a string that stopped being verbatim one line later. Measured before the second fix:

```
"\\?\%LOCALAPPDATA%\wenlan.\dev" -> "C:/Users/h1646/AppData/Local/wenlan./dev"
guard says within-production: false
```

and Win32, opening that, strips the dot and lands in `...\wenlan\dev`. The exemption had reopened the production bypass for one spelling. (Round 6.)

Carrying the prefix through instead is not available: MSYS `mkdir` and `realpath` and the daemon all go through Win32, and `realpathSync.native` drops it regardless. So neither reading of a verbatim path can be honoured, and the script now says so. `reject_verbatim_path` refuses one on all three dev inputs, before any canonicalization, with a message naming the fix. Windows only: a leading `\\` is an ordinary relative filename on POSIX. `canonicalize_path` goes back to stripping unconditionally, which is what Win32 does to every path that can still reach it.

Listing the prefixes was the same mistake one level down. Win32 accepts either separator in all three positions, and `path.resolve` folds all sixteen combinations - `\\?/`, `//?\`, `/\./` and the rest - into the same two canonical forms, so four literals left twelve ways in. (Round 7.) The guard matches the shape instead, `[\\/][\\/][?.][\\/]`, which is the rule rather than a sample of it. `\\server\share` is untouched: the third character has to be `?` or `.`.

The two tests round 5 added asserted on `print-config`, which prints the *uncanonicalized* value and so never exercised the redirect at all - a real gap, and the reason the round-6 defect survived a green suite. The replacements drive the top-level guard directly, and one of them is platform-neutral by construction: it reads the regex text out of `dev-runtime.sh`, hands it to `RegExp`, and asserts it agrees with `path.win32.resolve` about which inputs are device paths. `path.win32` does not need to be on win32, so app-check's macOS runner now constrains the shape even though it cannot run the guard.

Measured on this machine after all of it: `wenlan`, `WENLAN`, `wenlan.`, `wenlan.` with trailing spaces, `WENLAN./data`, `wenlan/`, `./wenlan`, `origin/../wenlan`, `OrIgIn/x`, `WeNlAn/deep/nest`, and both directions of the `PROGRA~1` alias all exit 2 with `refusing production`; all sixteen device-prefix spellings exit 2 with `refusing extended-length or device path`. A fresh `mktemp -d` path still starts, the printed config is byte-identical, and the live start/stop cycle is unchanged (daemon PID 12204, health ok, listener matching, port freed, second stop idempotent). Twenty-two Windows cases cover the behaviour, and four platform-neutral cases cover the branches on app-check's macOS runner, where they cannot execute.

## Verification (physical Windows 11 machine, x86_64-pc-windows-msvc)

- `pnpm install` - succeeds
- `pnpm test` - 175 files, 1898 passed, 29 skipped, 0 failed
- `cargo clippy --workspace --exclude wenlan-app --all-targets -- -D warnings` - clean
- `cargo test -p wenlan-core --lib repair` - 119 passed, 0 failed, 115 ignored. Before this PR the same selection failed 115 of them
- `cargo test -p wenlan-core --lib eval::report` - 28 passed, 0 failed
- `scripts/ci_test_plan.test.py` - 93 tests, OK
- `cargo test -p wenlan-core --lib drift_guard::` - 157 passed, 0 failed
- `pnpm vitest run scripts/dev-runtime.test.ts scripts/prepare-sidecars.test.ts` - 38 passed, 28 skipped, 0 failed
- production-root guard, on this machine: every spelling listed in section 11 exits 2 with `refusing production` or `refusing extended-length or device path`; a fresh `mktemp -d` path still exits 0
- identity-checked kill, against a live decoy: survives a non-matching path, survives a sibling name in the same directory, dies to its own full path, no-ops on a dead pid
- `dev-runtime.sh start` then `stop`: daemon healthy on the isolated port with the netstat listener matching the recorded pid, then process gone, port free, second `stop` idempotent
- `.claude/hooks/test-hooks.sh` - 8/8
- `pnpm dev:all` - runs end to end: daemon on the isolated port, vite on the isolated UI port, app window up, then a clean teardown
- `git push` - completes the pre-push closure and succeeds, which is what this PR is really about

The full `cargo test --workspace --lib` run is CI's (`workspace-lib-required`); the Windows receipt above covers every case that was failing.

Also verified on the same machine, unrelated to these fixes but worth recording: all four binaries build from source, `smoke-windows.ps1` passes (adjacent `vulkan-1.dll` and `onnxruntime.dll` load, store and semantic search work), and all three Qwen live smokes pass including GPU selection of an RTX 3060 and visible CPU fallback.

## The desktop app itself, on Windows

With defects 8 and 9 fixed, `pnpm dev:all` builds and runs the whole stack. Walked the entire first-run flow with real mouse and keyboard input against an isolated dev daemon (port 17895, scratch `WENLAN_DATA_DIR`):

- window opens, titled `Wenlan`, native title bar and taskbar entry
- setup wizard step 1 renders and `Get started` advances it
- step 2 reports real hardware read off this box: `Your system: 16GB RAM - CUDA GPU`, offering `Qwen 3 4B`
- step 4 detects the agents actually installed here - Claude Code and Codex CLI, both `CONFIGURED`
- step 5 runs the live runtime check and reports **"Wenlan is running - it stored a test memory and read it straight back"**, which is a real round trip through the daemon
- the wizard completes and the main app renders: sidebar (Home / Wiki / Spaces / Graph / Memories / Sources), search, Activity, and the dated greeting. Navigating to Memories works.

On the `dev:all` run the webview loads from the injected `devUrl` (`http://localhost:18895/`), reaches `http://127.0.0.1:17895/api/health` from inside the app, and the app logs `[daemon] Existing healthy daemon on port 17895 - exiting cleanly` - it attaches to the daemon the dev runtime started instead of spawning its own. On exit, `dev-all.sh`'s trap reports `Stopped worktree-owned Wenlan dev daemon (PID 2912)` and both ports are free.


7xuanlu:
Independent closure review (Codex `gpt-5.6-sol`, xhigh reasoning, read-only) returned **CODE READY: YES** on `d6adf8ac` — "No concrete defects found." That is pass 8; every finding from rounds 1-7 is fixed, tested and recorded in the PR body above. GitHub will not let the author self-approve, so this comment is the review record.

CI run [31941455814](https://github.com/7xuanlu/wenlan/actions/runs/31941455814) on the same head: success.

PR #588, opened 2026-08-24
ci(gauntlet): SIGKILL watchdog for post-mortem daemon replays

7xuanlu:
## What

The from-main gauntlet run (32706075944) confirmed F4 via the discriminating replay: with an absolute `FASTEMBED_CACHE_DIR` the daemon printed `WENLAN_LISTENING_ON=127.0.0.1:17918` under the exact launchd environment shape, while the default-cache replay died with the embedder error — same host, same minute, one variable. But the recovered daemon then outlived the replay's 30s cap: all three macOS jobs hung until their own `timeout-minutes`, GitHub reported them (and the run) as **cancelled**, and the `daemon-replay-cache-dir` verdict row was lost.

## Fix

- `scripts/first-run/lib.sh`: replace the `timeout`/`gtimeout` invocation with a shared `_replay_capped` helper — daemon in the background of the replay subshell, poll `GAUNTLET_REPLAY_CAP` seconds (default 30), then SIGKILL. A KILL watchdog cannot be ignored or waited out, so the info row is always written and the job never hangs.
- `scripts/first-run/lib.test.sh`: new case proves a TERM-ignoring daemon is killed with rc 124 (cap shortened to 2s for the test).

## Verification

- `bash scripts/first-run/lib.test.sh`: 38/38 assertions pass (4.5s), including the new watchdog case.
- `shellcheck -x` clean on both files; `bash -n` clean.

### DIFF
--- a/scripts/first-run/lib.sh
+++ b/scripts/first-run/lib.sh
@@ -186,15 +186,10 @@ daemon_postmortem() {
         return 0
     fi
     [ -x "$bin" ] || { info daemon-replay "skipped: $bin is not executable"; return 0; }
-    local replay="$GAUNTLET_OUT/checks/daemon-replay.log" rc=0 timeout_bin
-    timeout_bin="$(command -v timeout || command -v gtimeout || true)"
-    (
-        cd / && env -i HOME="$HOME" USER="${USER:-$(id -un)}" PATH=/usr/bin:/bin:/usr/sbin:/sbin \
-            RUST_LOG=info WENLAN_DATA_DIR="$data_root" WENLAN_BIND_ADDR=127.0.0.1:17917 \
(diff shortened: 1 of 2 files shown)


7xuanlu:
Correction from adversarial review round 2 (Codex Sol): this PR's diagnosis was wrong. The macos-15 runner image ships neither `timeout` nor `gtimeout`, so the pre-fix optional cap silently did not exist — the replay daemon was never signaled at all (the released daemon's TERM handler drains and `_exit`s in 1.5 s, so a working `timeout 30` would have ended it). The SIGKILL watchdog fix itself stands and removes the tool dependency. On-runner proof dispatched as run 32713127051; two further evidence-gap hardenings in the follow-up PR.

PR #590, opened 2026-08-24
fix(server): wire the four /api/pages/drafts editor routes

7xuanlu:
## Summary

Audit item `server#0` (deferred on purpose at v0.17.0): the desktop page editor's autosave/publish/discard chain calls POST `/api/pages/drafts`, PUT/DELETE `/api/pages/drafts/{id}`, and POST `/api/pages/drafts/{id}/publish`, but the daemon never registered those routes, so every draft save and publish failed. Decision: **wire** (the UI chain — `PageDraftEditor.tsx`, `usePageDraftAutosave.ts`, `app/src/search.rs` commands, wenlan-types wire shapes — is all live).

## What's here

- **wenlan-core** — new `publish_page_draft(id, expected_version)` with `PageDraftPublishOutcome` (Published / VersionConflict / TitleConflict). Semantics mirror the executable editor spec in `e2e/tauriMock/runtime.ts`: exact-retry replay (active row at `expected_version + 1`), version-conflict precedence over not-a-draft, trimmed-title + non-empty-content requirement, same-scope case-insensitive title conflict among active pages. Publish stamps the trimmed title, `status='active'`, `kind` via the shared `page_kind_for` rule, bumps the version, and computes the page embedding best-effort outside the write transaction (FTS reindex rides the existing `pages_fts_update` trigger).
- **wenlan-server** — four handlers on the existing `wenlan-types` request/response shapes (omitted `space` inherits `X-Wenlan-Space`; explicit `null` clears it). New `ServerError::Structured` passes route-owned status + JSON bodies through untouched, carrying the codes `src/lib/tauri.ts` `parsePageDraftError` parses: `draft_version_conflict` (+`current_version`), `page_title_conflict` (+`existing_page_id`/`existing_page_title`), `page_draft_not_found`, `page_draft_id_conflict` — messages verbatim from the mock. Publish writes the md projection through the same `lint_config.page_root()` as `handle_create_page` (drafts themselves are never projected).
- **Contracts** — handler manifest rows, `NON_SENSITIVE_PATHS` (pure write paths), truth-manifest rows (draft echoes are page-bearing, `MarkerShape::None`; DELETE has no prose fields) + count assertions, regenerated m5 reader inventory.
- **Tests** — core publish coverage in `page_drafts_test.rs` (flip/replay/conflict/validation/title-conflict incl. archived non-conflict); HTTP-surface contract in `crates/wenlan-server/tests/page_draft_routes.rs` (full lifecycle, error-code bodies, space-header semantics), per the audit's "test at the HTTP surface".

## Verification

- `cargo test -p wenlan-server --lib` 384 pass; `--tests` 644 pass (44 suites)
- `cargo test -p wenlan-core --lib page_drafts` 24 pass; `truth_manifest` 16 pass; `drift_guard` 156 pass; serving catalog 19 pass
- `cargo fmt --all -- --check` clean; `cargo clippy -p wenlan-core -p wenlan-server --all-targets -- -D warnings` clean


7xuanlu:
Adversarial review (Codex Sol, gpt-5.6-sol) of 77367a9 returned **DO NOT MERGE** with five P1 findings. All five verified against source and fixed in c076567e:

1. **Unicode title conflicts missed** — bundled SQLite `lower()` is ASCII-only. The publish conflict check now scans same-scope active pages and folds in Rust (`trim().to_lowercase()`), the same reasoning as migration 31. Test: `publish_folds_unicode_titles_when_checking_conflicts` (Greek/Cyrillic pair).
2. **page_history invariant violated** — publish now calls `append_page_history(&tx, id, "publish", ts)` inside the publish transaction (the documented ONE site for version bumps).
3. **Wikilink graph left stale** — publish runs `refresh_page_wikilinks` on the published body plus `resolve_orphan_page_links`, both post-commit and warn-on-failure, mirroring `insert_page`/`update_page_content`. Test: `publish_appends_history_and_maintains_wikilinks` (outgoing edge + orphan resolution both directions).
4. **Truth-manifest classification dishonest** — the three page-echo handlers now take the `TruthView` extractor and pass the echo through `truth_adapter::filter_page` (`filter_draft_echo`); a hidden echo answers with the structured 404. Safe for the editor: fresh drafts/publishes are `Unevaluated → Full`; only a replay of a since-classified-unsupported page hides. Manifest/inventory evidence updated. No end-to-end unsupported-replay HTTP fixture (filter semantics are unit-covered in `truth_contract_test.rs`); flagged for the re-review.
5. **Update/discard after publish returned 422** — `update_page_draft`/`delete_page_draft` now return the structured 404 `page_draft_not_found` for non-draft rows (publish keeps replay + validation semantics). Tests at both layers.

Sol's P2s (projection warn-only success, content canonicalization parity with the mock) are deliberate divergences left as follow-ups.

Local gates: core `page_drafts` 27, `truth_manifest` 16, `drift_guard` 156, `serving` 19, full server suite 648/45 suites, clippy `-D warnings` clean, fmt clean. Sol re-review of the delta is running; merge waits for it plus CI.

7xuanlu:
Second adversarial review round (Codex Sol) on the fix commit returned DO NOT MERGE with two new P1s, both fixed in f4ab213a:

**P1 — Unicode link identity vs conflict identity.** Publish folded titles in Rust while `find_unique_active_page_id_by_title_scoped` folded in bundled SQLite `LOWER()` (ASCII-only), so `[[i σχεδιο проект]]` stayed orphaned next to an active "I ΣΧΕΔΙΟ ПРОЕКТ" that the conflict check called a duplicate. Both now fold through one `page_title_key` seam; `wikilinks_fold_unicode_titles_like_the_conflict_check` covers publish-time resolution and orphan re-resolution across Greek/Cyrillic casing.

**P1 — TitleConflict identity leak + no load-bearing filter test.** The 409 named `existing_page_id`/`existing_page_title` even when the truth contract hides that page, and every route test ran at generation 0 where the adapters are pass-through. The handler now filters the conflicting page and omits both optional fields when hidden. Two live-cutover HTTP tests (raw-SQL Unsupported fixture, same seam as `refinery_routes.rs`) pin it: `publish_replay_of_a_hidden_page_reports_draft_not_found` and `title_conflict_with_a_hidden_page_omits_its_identity`, each with a visible-state control so the flip is provably the verdict's doing.

**P2 (follow-up, per reviewer)** — post-commit link refresh is not version-CAS-bound and a replay cannot repair a crashed refresh; inherited from the established best-effort seam shared with `insert_page`/`update_page_content`.

Verification: page_drafts 28, wikilink 27, orphan 66, HTTP draft routes 5, full wenlan-server suite 650 passed, drift guard 156 (inventory regenerated), truth manifest 16, clippy `-D warnings` and fmt clean.

7xuanlu:
Round 3 (`fb5aa856`) addresses the third-round review verdict (DO NOT MERGE, 2×P1 + 3×P2):

**P1 — resolver performance.** Title resolution now builds one folded-title ambiguity map per scope scan (`folded_title_owners_scoped`, no `ORDER BY`, no per-label row transfer). `find_unique_active_page_id_by_title_scoped`, wikilink resolution (`resolve_against_pages`), and the post-publish orphan sweep (one map per distinct scope) all consume the same map, so a publish resolves every label and orphan from at most one scan per scope.

**P1 — conflict disclosure not bound to the disclosed identity.** The 409 body now discloses identity via a new `visible_conflict_identity` helper: one reload of the row, filtered through the caller's truth grant, and disclosed only when the reloaded row still folds to the conflicting title. The disclosed id/title come from that reloaded row, never the transaction's capture — a deleted row, a reused id with an unrelated title, or any reload/filter failure degrades to omission while keeping the 409 (also resolving the P2 about 500s on identity-read failures). Four deterministic unit tests cover visible / deleted / replaced-unrelated / replaced-same-fold (concurrent replacement simulated via a second libsql connection, the established routes-test seam).

**P2 — stored-title trim untested.** New core test `find_unique_folds_untrimmed_stored_titles`: a padded stored title (untrimmed `insert_page` path) resolves for its trimmed fold, and a clean twin then makes the pair ambiguous.

**P2 — identity-domain comment.** The `page_title_key` doc now states the eligible row sets differ on purpose (publish conflict scan includes entity-shadow rows, link resolution excludes them); predicate alignment stays a follow-up.

Verification: `wenlan-server` lib 4/4 new + `page_draft_routes` 5/5; core slices page_drafts 28, wikilink 27, orphan 66, find_unique 2, drift_guard 156 (reader inventory regenerated); clippy `-D warnings` clean for both crates incl. `--features eval-harness`; fmt applied.

7xuanlu:
Round 4 (`343b45ad`) addresses the fourth-round verdict (DO NOT MERGE, 3×P1). All three shared one mechanism failure — identity decisions made outside the snapshot they act on — so this round moves them onto the daemon's real concurrency boundary, the single connection mutex:

**P1 — TOCTOU in conflict disclosure.** New core method `MemoryDB::conflict_identity_snapshot(grant, page_id, wanted_key)`: one connection guard covers the row read, the truth verdict (via a private `page_truth_states_on_conn` extraction, whose SQL already joins the live `pages` row), and the title-fold recheck. Every writer goes through the same mutex, so the rename interleaving cannot occur — the disclosed identity and the visibility decision are one snapshot by construction, not by test coverage. The server helper is now a thin wrapper degrading any error to omission.

**P1 — stale owner map in the orphan sweep.** `resolve_orphan_page_links` now runs the entire sweep — orphan list, per-scope folded-title maps, deletes, edge mints (including the target-space read) — inside one transaction on one connection guard. The snapshot-faithfulness comment is now true by construction; one failed row rolls back the whole sweep, which is an idempotent repair the next publish re-runs.

**P1 — single-label cost.** The one-label path no longer builds or retains the owner map: `find_unique_active_page_id_by_title_scoped` fold-compares each row and stops at the second match (the old `LIMIT 2` shape), and the wikilink resolver routes single-label sets there. Measured receipt (new ignored test `find_unique_scan_cost_receipt`, release build, 2,000-page scope, 100 lookups): **740µs/op for the Unicode-correct Rust fold vs 371µs/op for the ASCII-only `LOWER()` SQL it replaced**. The remaining 2× is the irreducible price of a correct fold without an indexed folded column — SQLite's `LOWER()` is ASCII-only, so any correct implementation must transfer the titles. At sub-millisecond absolute cost on a scope larger than typical, we judged a folded-column schema migration (touching every `pages` insert/update site) not worth its risk; the receipt is in-tree to re-run if scopes grow.

Verification: drift_guard 156 (R4 raw-handle guard required the `_on_conn` extractions to be private; reader inventory regenerated, 402 rows), orphan 66, wikilink 27, page_drafts 28, truth_exposure 39, find_unique 2, server `conflict_identity_tests` 4/4 and `page_draft_routes` 5/5, clippy `-D warnings` clean on default and `--features eval-harness`, fmt applied.

7xuanlu:
Round 5 (`a07f6d26`) — response to Sol's fifth-round DO NOT MERGE (two P1s, three P2s).

**P1 — conflict snapshot now revalidates the full predicate.** `TitleConflict` carries the stored (sentinel-mirrored) scope the publish transaction found the conflict in, and `conflict_identity_snapshot` requires `space = ?2 AND COALESCE(kind,'concept') != 'entity'` alongside id + active. The kind fence restores the entity-shadow omission the old `get_page` path enforced. New regressions: `a_page_moved_to_another_space_degrades_to_omission`, `an_entity_shadow_conflict_stays_omitted`.

**P1 — orphan sweep no longer deadlocks on a poisoned row.** Each orphan repairs inside `SAVEPOINT orphan_repair`; a row failure rolls back alone (a failed rollback aborts the sweep so partial writes can never commit), the identity is logged at warn, and the sweep continues. New regression `a_poisoned_orphan_does_not_block_the_other_repairs`: a pre-existing edge row with payload `'not json'` under the exact id the sweep upserts, bad source ordered first — the good orphan resolves, only the poisoned one remains, and a second sweep still returns Ok.

**P2s**: the orphan list is capped at `ORPHAN_SWEEP_BATCH = 512` rows per invocation (warn on cap hit, remainder drains next sweep); the snapshot's truth filter is now `generation != 0`, matching canonical fail-closed behavior; the stale "drops the connection" comment is rewritten and `bench-out.txt` is deleted.

**Verification at `a07f6d26`**: full core lib suite 4096 passed / 0 failed; server suite 390 passed / 0 failed; clippy `-D warnings` clean (core default + eval-harness, server); conflict_identity (6), page_draft (31), resolve_orphan (2), drift_guard (156), truth (115) slices green; m5 reader inventory regenerated (402 rows). Sol round 6 running.

7xuanlu:
## Round 6 — response to Sol's sixth review (9e4035b8 + 3349b2f3, head 3349b2f3)

Sol's round-6 verdict was DO NOT MERGE on three points.

| Finding | Response |
|---|---|
| P1 — the orphan sweep always re-selected the same ordered prefix of 512 rows, so 512 unresolvable or poisoned rows in front hid every later row from every future sweep | `resolve_orphan_page_links` now keeps a durable keyset cursor (`orphan_sweep_cursor` in `app_metadata`, a JSON pair of `source_page_id` and `label_key`). It fetches one row beyond the batch so the "more remain" warning is truthful, advances past every examined row (resolved, unresolvable, or failed), wraps to the head after the tail, and clears the cursor there. The cursor rides the sweep's own transaction, so a rolled-back sweep never advances it. The warning now fires only after the cursor is saved and names the cursor pair. |
| P2 — the `generation != 0` fix had no regression | `a_negative_generation_still_hides_an_unsupported_conflict_identity` sets the cutover generation to -1 and asserts the provisional page's identity is omitted while the supported page's is still disclosed. |
| The SIGABRT core run | That was a concurrent run on a shared build directory; the solo rerun of a07f6d26 passed 4096 tests with exit code 0. |

Regressions for the cursor: `a_full_batch_of_unresolvable_orphans_does_not_starve_the_tail` puts 512 unresolvable links and one resolvable link on a single source page, so the cursor must advance mid-page (sweep 1 returns 0 and sets the cursor, sweep 2 returns 1, sweep 3 returns 0 and clears it); `a_stale_cursor_past_the_tail_falls_back_to_the_head` covers a cursor written past the tail.

Mutation controls run by hand, each restored afterwards: ignoring the cursor fails sweep 2 (`left: 0, right: 1`); deleting the intra-page arm of the keyset predicate fails sweep 2 the same way; reverting `!= 0` to `> 0` fails the generation test (`left: Some(("p2", "p2")), right: None`).

An independent review of the diff (correctness, transaction semantics, keyset ordering under the real schema, index use) returned SHIP after its five findings were applied in 3349b2f3.

Behaviour note: while more than 512 orphan rows are pending, a newly resolvable row behind the cursor waits for the wrap — at most ceil(orphans / 512) publishes — instead of never resolving; databases with 512 or fewer orphan rows never write the cursor and are unaffected.

Out of scope, flagged for a separate ticket: if the sweep's final `COMMIT` fails, the error returns without a `ROLLBACK`, leaving the transaction open on the shared connection. This predates the PR (identical at a07f6d26).

Receipts at 3349b2f3, each run alone on the build directory: `wenlan-core` lib suite 4098 passed, 0 failed, 32 ignored, exit code 0; `wenlan-server` suite every binary 0 failed, exit code 0. CI: 15 passed, 0 failed.

Sol's seventh review round is running against this head. The branch is behind main after #591 merged; a dry-run merge shows no conflicts, and the branch will be updated before merging.

7xuanlu:
Merging. Sol's seventh review round (head `3349b2f3`) returned **MERGE** with no new blockers; per the agreed stop rule that was the final round.

Gates at the integrated head `90737cbc` (branch merged with `main` after #591 landed, no conflicts):

- CI: 15 passed, 0 failed, 10 skipped; merge state CLEAN.
- Local focused core tests (`orphan`, `conflict_identity`, `negative_generation`): 69 passed, 0 failed.
- Local server draft-route tests: 3 passed, 0 failed.
- Full core and server suites were green at `3349b2f3` (4098 passed / 0 failed; server binaries 0 failed).

Out of scope, filed separately: #593 (a failed `COMMIT` in `resolve_orphan_page_links` leaves the writer transaction open).

PR #591, opened 2026-08-24
fix(core): surface row-scan step errors instead of silently truncating

7xuanlu:
Audit item core-db#7: 57 row loops used `while let Ok(Some(row)) = rows.next().await`, which turns a mid-scan step error into a silent early stop — including migration 43's pragma scan, which then reported success over a truncated read.

## What changed

- **41 sites in `Result` fns now propagate** the step error via `map_err` + `?` with per-fn context, following the existing migration-40 idiom (`db.rs`, `db/scoped_entities.rs`, `db/eval_lifecycle_integrity.rs`).
- **16 sites inside deliberately best-effort retrieval channels** (vector/FTS fallback arms that already warn-and-continue on query failure: `search`, `search_memory_with_cue`, `search_episodes_scoped`, `search_facts_channel`, `naive_vector_search`, `fts_only_search`, `search_entities_by_vector`, `search_summary_nodes_scoped`, `search_pages`) now `log::warn!` the aborted scan and stop, instead of stopping silently. Propagating there would change the channels' intended degraded-mode behavior.
- The 5 remaining grep hits live in `eval/shared.rs` `mod tests` (test support, non-Result fn) and are out of audit scope.

## Test

`supersedes_scan_surfaces_mid_scan_step_errors`: proves the healthy path first (2 rows), then swaps the `memories` table for a view whose column expression raises at one row (`json_extract` over malformed JSON — a real SQLite step error mid-scan) and asserts `eval_lifecycle_supersedes_inputs` now returns the contextual error instead of silently truncated rows. The migration-43 pragma loop itself is not fixture-poisonable (PRAGMA output), so the mechanism test stands in for it.

## Verification

- `cargo clippy -p wenlan-core --all-targets -- -D warnings` clean
- full `wenlan-core` lib suite: 4124 passed
- `cargo fmt` clean

### DIFF
--- a/crates/wenlan-core/src/db.rs
+++ b/crates/wenlan-core/src/db.rs
@@ -28742,7 +28742,20 @@ impl MemoryDB {
             .map(|(rank, id)| (id.clone(), rank))
             .collect();
         let seeds = crate::retrieval::signals::seed_entities_by_rank(&pool_pairs, top_k);
-        let seeds = self.filter_entity_ids_scoped(&seeds, scope).await?;
+        // Graph augmentation is optional: propagating this error would drop
+        // `results` (already moved here), and the caller's retry-with-empty
+        // path would then erase healthy base hits. Degrade to the base
+        // results instead; the filter's own decode errors stay loud for
+        // callers that ask it directly.
+        let seeds = match self.filter_entity_ids_scoped(&seeds, scope).await {
+            Ok(seeds) => seeds,
+            Err(e) => {
+                log::warn!(
+                    "[augment_seeded] scoped seed filter failed; graph contributes nothing: {e}"
+                );
+                return Ok(results);
+            }
+        };
 
         // BFS to expand seed set
         let expanded = match self
@@ -28970,14 +28983,18 @@ impl MemoryDB {
                 params.push(dp.clone());
             }
 
-            match conn.query(&fts_sql, params).await {
-                Ok(mut rows) => {
-                    // Eval baseline: a mid-scan step or decode error is
-                    // silent truncation of the scored candidate list, so it
-                    // propagates. A query error is lenient only while a
-                    // fallback form remains — the AND form can be invalid
-                    // FTS5 syntax for free text and falls through to the OR
-                    // form, but the final attempt failing is a real error.
+            // Eval baseline: silent truncation of the scored candidate list is
+            // the failure this exists to remove, so the WHOLE attempt — query,
+            // scan, rank, and decode — is fallible as one unit. A failure on a
+            // non-final form (the AND form can be invalid FTS5 syntax for free
+            // text) discards that attempt's partial prefix and falls through
+            // to the next form; the final form's failure is the caller's error.
+            let attempt_result: Result<Vec<SearchResult>, WenlanError> =
+                async {
+                    let mut attempt_results = Vec::new();
+                    let mut rows = conn.query(&fts_sql, params).await.map_err(|e| {
+                        WenlanError::VectorDb(format!("fts_only_search query: {e}"))
+                    })?;
                     while let Some(row) = rows.next().await.map_err(|e| {
                         WenlanError::VectorDb(format!("fts_only_search row scan: {e}"))
                     })? {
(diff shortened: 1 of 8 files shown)


7xuanlu:
Second sweep (`36483aa7`) addresses the adversarial review of `287b866f` (MERGE WITH FIXES, 3×P1 + 2×P2):

**P1 — 18 remaining production swallows.** The first sweep's pattern missed row reads shaped as `if let Ok(Some(...))`. All are now propagated with contexts: `pipeline_status` (entity row + raw count), `list_recent_retrievals` (snippets + resolve), `get_enrichment_summary`, `get_memory_source_agent`, `get_memory_entity_id`, the entity-enrichment vector resolve, `resolve_entity_by_name` (fuzzy + exact), `update_document_content`, `get_memory_details`, `count_unembedded_chunks`, the four `eval_lifecycle_integrity` count scans, and two wrong context strings in `scoped_entities.rs`. The `eval/shared.rs` hits are inside `#[cfg(test)]` and out of scope.

**P1 — graph augmentation hard-failing primary search.** `augment_with_memory_stream` (the `WENLAN_GRAPH_MEMORY_STREAM` channel, default on) now catches both the anchor-eligibility and linked-memory fetch errors, warns, and returns the primary results unchanged — the graph channel degrades instead of taking search down with it. `search_summary_nodes_scoped`'s root lookup likewise warns and aborts cleanly rather than scoring against a partial root set.

**P1 — eval baselines must not truncate silently.** `naive_vector_search` propagates query and row-scan errors; `fts_only_search` propagates row-scan errors while keeping the AND-query error lenient by design (invalid FTS5 syntax falls to the documented OR fallback). Eval runners (`context_path.rs`, `answer_quality.rs`) now call a new strict `search_pages` variant that propagates scan errors, while production `search_pages` stays best-effort.

**P2 — DiskANN partial prefix.** A primary vector-scan failure now sets a flag that triggers the brute-force fallback even when partial results came back, instead of gating the fallback on `results.is_empty()`.

**P2 — wrong contexts.** Both `scoped_entities.rs` strings now name their actual functions.

Verification: full `wenlan-core` lib suite 4088 passed / 0 failed (exit code 0); clippy `-D warnings` clean on the default features and with `--features eval-harness`; targeted slices (eval_lifecycle 4, search 60) green; fmt applied.

7xuanlu:
Round 3 (`1813c4dd`) — response to Sol's second-round MERGE WITH FIXES verdict.

**All five P1s fixed** — every flagged silent decode now propagates:
- legacy-chunks count, `get_table_columns`, m113/m114 scan ids, `archived_ids`, `filter_entity_ids_scoped`, eval-lifecycle merged ids, degree-stats count
- `naive_vector_search` distance + row decode
- `fts_only_search`: non-final attempts warn and fall through; the final attempt's failure returns `Err`
- `search_summary_nodes_scoped`: any root query/scan/decode failure empties the whole summary channel (root-first invariant)
- `resolve_entity_by_name` step C and the LongMemEval caller: a failed question now fails the run instead of silently scoring a biased partial sample

**P2s**: `augment_with_graph_gated` logs the entity-anchor failure instead of dropping it; two new failure-path regression tests:
- `a_failed_summary_root_query_empties_the_channel` — renames `summary_nodes` via the opaque secondary test session, asserts `Ok(empty)` after proving the healthy path returns the root
- `strict_page_search_fails_on_a_row_the_lenient_path_skips` — desyncs the FTS5 shadow table so `MATCH` fails at step time; strict returns `Err("… fts row scan …")`, lenient still returns the page via the vector arm

Injection note: every hard-decoded `pages` column is `NOT NULL` and libsql's `FromValue` panics (`unreachable!`) on stored type mismatches rather than erroring, and the DiskANN index rejects broken embedding writes — the FTS desync was the one clean seam for the mid-scan branch.

**Verification**: full core lib suite green at `1813c4dd` — 4090 passed, 0 failed. Clippy `-D warnings` clean (default + `eval-harness`); search/eval_lifecycle/scoped_entities slices green. Sol round 3 running.

7xuanlu:
## Round 4 — response to Sol's third review (e6989394)

Sol's round-3 verdict was MERGE WITH FIXES: four P1s and one P2. All five are addressed in e6989394.

| Finding | Fix |
|---|---|
| P1 `memories_entities_degree_stats` — the `memories_linked` count (the site flagged, not the grouped count fixed in round 3) defaulted to 0 on failure | The query error, the `next()` error, and the decode error now all propagate as `WenlanError::VectorDb` |
| P1 `fts_only_search` — a decode failure inside one FTS form could leak a partial prefix into the fallback | Each attempt (query, scan, rank, row decode) is one fallible unit; a non-final form's failure discards that attempt, warns, and falls through; only the final form's failure returns `Err` |
| P1 `row_to_summary_node` — `source_count` and `generated_at` were zeroed on decode failure | Both decode errors propagate, so a malformed root hits the root-decode early return and empties the channel |
| P1 seeded graph augmentation — a scoped seed-filter failure escaped with `?` and the caller's retry-with-empty erased healthy hits | The failure warns and returns the base results unchanged; `filter_entity_ids_scoped`'s direct callers still see propagated errors |
| P2 the summary-root regression could pass for a wrong reason | Rebuilt as `a_root_that_fails_to_decode_empties_the_channel` (`db/scoped_pages_test.rs`): Space scope, one root plus one healthy bucket, control asserts both are returned, then only the level-1 row's `generated_at` is nulled so an empty result can only come from the root-decode branch |

Mutation control run by hand on the rebuilt test: reverting the `generated_at` hard-decode makes it fail on "not degrade to buckets".

Receipts at e6989394: clippy `-D warnings` clean (default and eval-harness feature sets); full `wenlan-core` lib suite 4090 passed, 0 failed, 31 ignored, exit code 0 (run alone on its own build directory).

Sol's fourth review round is running against this commit.

7xuanlu:
Sol's fourth review round at e6989394 returned **MERGE** with no remaining findings: all four round-3 P1s verified in code, the rebuilt summary-root regression judged valid (neither a zeroed root nor a skipped root can satisfy the empty-result assertion), and the full-suite receipt confirmed (4090 passed, 0 failed, 31 ignored, rc=0). CI at this head: 14 passed, 0 failed.

Squash-merging now; PR #590 will rebase onto the result before its own merge.

Issue #586, opened 2026-08-24
fix(server): blocked agent storing a memory gets HTTP 500 instead of 403


7xuanlu:
## What happens

When an agent that has been disabled (blocked) calls `POST /api/memory/store`, the daemon answers HTTP 500 "Internal error" instead of HTTP 403 "Forbidden". The agent cannot tell that it was blocked on purpose; the response reads like a daemon fault.

## Where (read on `main` at 8b0a7ade, 2026-08-23)

`crates/wenlan-server/src/memory_routes.rs:384-386`:

```rust
db.check_agent_for_write(&resolved_agent)
    .await
    .map_err(|e| ServerError::Internal(e.to_string()))?
```

`check_agent_for_write` returns `WenlanError::AgentDisabled` for a blocked agent. The server already knows how to map that: `crates/wenlan-server/src/error.rs:96` converts `WenlanError::AgentDisabled` into `ServerError::AgentDisabled`, and `error.rs:78` renders that as `StatusCode::FORBIDDEN`. The store handler bypasses both by forcing every error through `ServerError::Internal`.

## Fix

Replace the closure with the existing conversion at the same line:

```rust
.map_err(ServerError::from)?
```

Add one server test: register an agent, disable it, store as that agent, assert 403 and the "Agent disabled" message. Check the other write handlers that call `check_agent_for_write` for the same pattern.

## Origin

Found while writing the spec for proposal-gated agent writes (`~/.wenlan/specs/wenlan/2026-08-23-proposal-gated-agent-writes.md`, open finding 8). Kept out of that PR on purpose so it stays a one-purpose change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01WiKD4Z76b9Nhvgm9robUj2

PR #592, opened 2026-08-24
feat(server,core): stage an agent's correction for review when its trust is not full

7xuanlu:
## What changes

Memory A says the meeting is on Tuesday. An agent whose trust level is not `full` sends a correction saying Wednesday, with `supersedes` pointing at A. Before this change the correction replaced A at once. Now A stays live and searchable, and the correction waits in the review list that already exists. Accepting it hides A and makes the correction live. New memories, which carry no `supersedes`, are unaffected and land immediately.

A local first-party write with no agent identity still counts as full trust and is never gated.

## Why

This is the smallest shape of "agent writes as reviewable proposals", borrowed from Tana and SiYuan during the 2026-08-16 feature study. It reuses the human-owned-page review card path rather than adding a proposal table, a new tool, or a new screen, in line with the standing "delete surface, do not add" decision. The design brief weighed three options and picked this one; all seven of its open questions were answered with the defaults.

## The rule

`pending_revision = req.supersedes.is_some() && trust_level != "full"`, at one site in `crates/wenlan-server/src/memory_routes.rs`. The flag then flows through both the ingest batcher and the fallback write path untouched.

## Also fixed here

Three distillation pool queries in `crates/wenlan-core/src/db.rs` did not exclude rows with `pending_revision = 1`, while their sibling `load_summary_buckets` did. Without that exclusion a staged correction could reach a distilled page before anyone reviewed it. This change would widen that pre-existing leak, so it is closed in the same commit.

## What needed no new code

Accept and dismiss already handle a memory target and fall through to `db.accept_pending_revision` and `db.dismiss_pending_revision`. The pending list, the `wenlan curate` command, and the desktop review queue all map the staged row straight through.

## Surfaces

The store response gains `gated: true`, matching the existing page write response. The MCP capture tool prints a line saying the correction is staged for review rather than live, so an agent reading only the success text cannot believe its correction took effect.

## Verification

- `cargo test -p wenlan-server`: 643 passed, 3 ignored, 0 failed.
- `cargo clippy -p wenlan-types -p wenlan-core -p wenlan-server -p wenlan-mcp --all-targets -- -D warnings`: clean.
- `cargo fmt --all -- --check`: clean.
- Core, MCP, and types suites plus a live proof on an isolated daemon are running; results will be added as a comment before merge.

## Out of scope

A blocked agent storing a memory currently gets HTTP 500 instead of 403. That is one line in the same file and is tracked separately as #586, kept out so this stays a one-purpose change.

Spec: `~/.wenlan/specs/wenlan/2026-08-23-proposal-gated-agent-writes.md`

https://claude.ai/code/session_01WiKD4Z76b9Nhvgm9robUj2


7xuanlu:
## Verification receipts

### Security escalation found in review, fixed in `965e18e3`

A pending revision whose `structured_fields` carry `revision_kind: "page_write"` and
`target_kind: "page"` is a **page revision card**: accepting it rewrites a page's prose,
dismissing it deletes the row. `structured_fields` is persisted verbatim from
`POST /api/memory/store`, and `resolve_page_revision_card` routed on those strings alone —
so a low-trust agent could stage a "memory correction" carrying the page markers and turn
the human's accept click into an overwrite of a human-authored page.

Two independent layers, each mutation-proven:

**Layer 1 — wire guard** (`crates/wenlan-core/src/origin.rs`). New
`strip_wire_page_revision_marker` removes the four card keys from any wire-supplied
`structured_fields` that claims them, keeping the caller's other fields, and logs the
dropped claim. Same shape as the existing `normalize_wire_source_agent` origin guard;
wired into the store handler in `memory_routes.rs`.

**Layer 2 — DB-proven page branch** (`crates/wenlan-core/src/post_write/page_revision.rs`).
The forgeable `revises_page` fallback is gone. The page branch now requires
`db.get_page(&payload.supersedes)` to resolve; a card whose `supersedes` is not a real page
is never treated as a page write. The one real minter, `stage_page_revision_card`, always
sets `supersedes` to the page id, so a genuine card is unaffected. This is also the fact
`list_pending_revisions_scoped` uses to label the card, so accept and the queue now agree.

### Mutation proofs

| Layer | Mutation | Result |
|---|---|---|
| 1 (wire guard) | made `strip_wire_page_revision_marker` a pass-through | `gated_store_cannot_forge_a_page_revision_card` FAILED (404 vs 200) |
| 2 (DB-proven branch) | restored the `revises_page` fallback | `a_card_whose_supersedes_is_not_a_page_is_never_a_page_write` FAILED (`left: "page_forged_victim", right: "mem_forge_target"`) |

The first attempt at proving layer 2 gave a **false negative**: the server-level test still
passed, because layer 1 had already stripped the markers before the row was written. The
core-level test was added specifically to isolate layer 2, seeding the forged
`structured_fields` directly with an `UPDATE` so the wire guard is out of the path.

### Local suites (all green)

```
cargo test -p wenlan-core --lib post_write   ->  117 passed, 4008 filtered out
cargo test -p wenlan-server --lib gated_store ->    6 passed,  379 filtered out
cargo test -p wenlan-core --lib origin::      ->    9 passed, 4116 filtered out
cargo test -p wenlan-core --lib r4_test_support_test -> 34 passed, 4091 filtered out
cargo fmt --all -- --check                    ->  clean
```

### CI fixes in `8a2841a2`

- **R4 drift guard** (`linux-nextest slice 3/3`): the test-support census is a hand-pinned
  exact count, not a generated file. The two new tests add 6 support calls, so the count
  moves 1027 -> 1033 and 6 rows join `r4_test_support_api_manifest.txt`. Rationale is
  written into the assertion message the way every prior delta is.
- **docs provenance**: the branch edits the Reconcile row in `README.md` and all three
  translations, but the `README_SYNC` hash markers still pointed at the pre-edit English
  body. Re-stamped with `python3 scripts/check-readme-translations.py --write-markers`.

### Not covered here

- Live proof against an isolated daemon (port + `WENLAN_DATA_DIR` + a scratch
  `knowledge_path`) has not been run; the guard is proven by unit and route tests only.
- The blocked-agent HTTP 500-vs-403 bug is deliberately out of this PR and tracked as #586.

PR #595, opened 2026-08-25
fix(core): split oversized distill clusters in steps instead of dropping them

7xuanlu:
## Summary

Top gap from the usefulness soak (`wenlan-usefulness-soak-2026-08-25.md`, gap 1): on a 28-memory store about one project, `/api/distill` returned no clusters and the plugin reported "up to date". The daemon log shows why:

```
[distill] found 26 eligible memories for clustering (raw excluded)
[distill] re-splitting oversized space cluster: 24 memories at threshold 0.65 (cap = 12)
[distill] dropping space sub-cluster after re-split: 20 memories still > cap 12 -- raise max_space_cluster_size if this is a genuine page
[distill] space_groups=1, unlinked=2, clusters_found=0
```

`cluster_distillation_rows` re-split an oversized group exactly once at `threshold + 0.05` and dropped whatever was still over the cap. A first-hour user who captures ~20 memories about one project never gets a page, and nothing outside the daemon log says so.

## Changes

- **core**: re-split in `+0.05` steps up to `0.92` before dropping (`RESPLIT_STEP`, `RESPLIT_MAX_THRESHOLD`). Only the branch that dropped clusters before is affected; groups that fit the cap today are clustered exactly as before. Tuning doc comment updated (it also described a stale 0.73 default; the code default is 0.60).
- **server**: `/api/distill` returns `clusters_found` (count before the covered-page filter) and a `hint` when nothing formed, so the caller can tell "nothing to cluster yet" from "everything is already covered".
- **plugin**: the distill skill's empty branch reports "No page formed yet" with the hint when `clusters_found` is 0, and keeps "up to date" only when clusters existed and were all covered.
- **test**: `oversized_space_cluster_resplits_in_steps_instead_of_dropping` — 4 topics × 5 memories in one space with cross-topic cosine ≈ 0.675 (survives the first re-split at 0.65, separates at 0.70). Written first: fails on the old code (0 clusters), passes with the fix. Outputs in the first comment.

## Verification

- Red/green run of the new test, the existing cluster tests, `cargo check` on server/mcp/types, clippy — receipts in the first comment.
- Real-surface proof: the fixed daemon run on a copy of the soak store returns page-sized clusters where 0.17.0 returned none — receipt in the first comment.

### DIFF
--- a/crates/wenlan-core/src/db.rs
+++ b/crates/wenlan-core/src/db.rs
@@ -2552,6 +2552,13 @@ fn has_capture_seed_floor(memories: &[ClusterMemRow], cluster: &DistillationClus
         >= MIN_CAPTURE_SEED_MEMBERS
 }
 
+/// Bounded re-split ladder for an oversized cluster (see `cluster_distillation_rows`):
+/// each round tightens the similarity threshold by this much...
+const RESPLIT_STEP: f64 = 0.05;
+/// ...until it hits this ceiling, past which an oversized sub-cluster is dropped
+/// rather than tightened further.
+const RESPLIT_MAX_THRESHOLD: f64 = 0.92;
+
 #[derive(Debug, Clone, Copy)]
 enum DistillationClusterMode {
     SpaceScoped,
@@ -2583,13 +2590,38 @@ fn cluster_distillation_rows(
                           indices: &[usize],
                           cap: usize,
                           clusters: &mut Vec<DistillationCluster>| {
-        let sub = cluster_by_similarity(memories, indices, similarity_threshold);
-        for group in sub {
+        // FIFO queue of (group, threshold-that-produced-it), seeded at
+        // `similarity_threshold`. An oversized group is re-split in
+        // `RESPLIT_STEP` increments (pushed to the back) rather than given a
+        // single re-split attempt, so a stubborn cluster tightens repeatedly
+        // up to `RESPLIT_MAX_THRESHOLD` before it is dropped. Popping from
+        // the front preserves today's output order for groups that never
+        // need a re-split.
+        let mut queue: std::collections::VecDeque<(Vec<usize>, f64)> =
+            cluster_by_similarity(memories, indices, similarity_threshold)
+                .into_iter()
+                .map(|group| (group, similarity_threshold))
+                .collect();
+
+        while let Some((group, threshold)) = queue.pop_front() {
             if group.len() < min_size {
                 continue;
             }
             if group.len() > cap {
-                let tighter = (similarity_threshold + 0.05).min(0.92);
+                if threshold >= RESPLIT_MAX_THRESHOLD {
+                    log::warn!(
+                        "[distill] dropping {} sub-cluster after re-split: \
+                             {} memories still > cap {} at threshold {:.2} -- raise \
+                             max_{}_cluster_size if this is a genuine page",
+                        bucket_label,
+                        group.len(),
+                        cap,
+                        threshold,
+                        bucket_label,
+                    );
+                    continue;
+                }
+                let tighter = (threshold + RESPLIT_STEP).min(RESPLIT_MAX_THRESHOLD);
                 log::info!(
                     "[distill] re-splitting oversized {} cluster: \
                          {} memories at threshold {:.2} (cap = {})",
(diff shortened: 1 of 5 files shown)


7xuanlu:
Receipts for `0749634a` (rebased onto main `57093695`, clean rebase; everything below ran on the integrated commit unless marked).

**Regression test, red then green** (run by the implementer on the pre-fix code, then with the fix):

```
thread 'db::tests::oversized_space_cluster_resplits_in_steps_instead_of_dropping' panicked at crates/wenlan-core/src/db/main_tests.rs:21698:5:
expected at least 4 clusters, got 0: []
test result: FAILED. 0 passed; 1 failed
```
```
test db::tests::oversized_space_cluster_resplits_in_steps_instead_of_dropping ... ok
test result: ok. 1 passed; 0 failed
```
Fixture self-check inside the test: cross-topic cosine 0.670 (inside the (0.65, 0.70) band the test depends on), within-topic 0.985.

**Cluster tests on the integrated commit**: `cargo test -p wenlan-core --lib -- oversized_space_cluster find_distillation_clusters sub_cluster_by_tokens` → `test result: ok. 12 passed; 0 failed` (includes `find_distillation_clusters_caps_oversized_entity_group` and `…_caps_oversized_unlinked`).

**Checks**: `cargo check -p wenlan-server -p wenlan-mcp -p wenlan-types` clean; `cargo clippy -p wenlan-core -p wenlan-server -p wenlan-mcp --lib -- -D warnings` clean; `cargo fmt --all` touched only the changed files; `git diff --check` clean. No typed distill response exists in `wenlan-types`/`wenlan-mcp` (`distill_impl` passes the daemon JSON through), so the two new fields reach the plugin as-is.

**Real-surface proof**: `wenlan-server` built from this commit, run alone on port 17918 with a fake `HOME`, `WENLAN_NO_AUTOSTART=1`, and a copy of the soak store (28 memories about one project in space `wenlan`, no model source). `POST /api/distill {}`:

| daemon | log | response |
|---|---|---|
| v0.17.0 (release) | `found 26 eligible` → `re-splitting … 24 memories at threshold 0.65 (cap = 12)` → `dropping space sub-cluster after re-split: 20 memories still > cap 12` → `clusters_found=0` | `pending: []`, plugin says "up to date" |
| this commit | `found 27 eligible` → `re-splitting … 25 memories at threshold 0.65` → `re-splitting … 16 memories at threshold 0.70` → `clusters_found=2` | `clusters_found: 2`, `pending: 2` (sizes 5 and 4, both space `wenlan`), no `hint` |

Honest limit: 9 of the 27 eligible memories reach a page-sized cluster; the rest split into groups under the 3-member floor at 0.70 and stay in the pool for later runs. That is the intended shape (topic-sized pages), not a drop.

7xuanlu:
## Review fixes (commit `bdbfcb7c`, rebased onto main `b49cc70d`)

The one adversarial review pass (Codex Sol, read-only, on `0749634a`) returned **MERGE** — no P1 — with three P2s and two P3s. Applied here:

| Finding | Fix |
|---|---|
| P2: one FIFO across all base groups queued re-split children behind every later sibling; with the stable size sort and `max_clusters` truncation, a tied cluster could swap which one survives (`[13, 5] → [5, 8]`, `max_clusters = 2`) | Each base group now drains with its descendants before the next sibling — the emission order the one-shot re-split had. Regression test `resplit_children_keep_their_base_group_position_for_max_clusters_ties` (the exact `[13, 5]` fixture; expects `[(8, a1), (5, a0)]`). |
| P2: the tuning loader never range-checks `formation_threshold`; `-inf`/NaN/huge negative makes `threshold + 0.05` stall and the oversized group was re-queued forever | The loop drops a group whose next threshold cannot advance (`threshold.is_finite() && tighter > threshold`). Regression test `a_threshold_that_cannot_advance_drops_instead_of_looping` (`-inf`, NaN, `-1e17`). |
| P3: the empty-pass hint blamed "in one space" (unlinked memories form pages too; a ceiling drop is not cured by capturing more) | Neutral text: *No page-sized cluster formed in this scope: nothing grouped into 3 or more related memories that fit one page. Capture more related memories, or check the daemon log for dropped clusters.* |
| P3: skill described `unresolved + hint` as JSON while the MCP tool renders it as prose | Skill now says the tool returns one line of text, ``Could not resolve target `<target>`. <hint>``, and to relay it verbatim. |
| P2: same-topic sibling shards can become several unconfirmed pages (no sibling-level dedup; LLM `split` action silently kept) | Not in this PR — tracked as #596 with the review's evidence and the two candidate fixes. |

Not done: load-time validation of `formation_threshold` and typed drop reasons with per-case route tests (both noted in #596).

### Mutation control (tests against the previous commit `ff401687`'s `process_bucket`)

```
test db::tests::resplit_children_keep_their_base_group_position_for_max_clusters_ties ... FAILED
  left: [(8, "a1"), (5, "b")]
 right: [(8, "a1"), (5, "a0")]
rc_order=101
inf_test: still running after 90 s -- loop confirmed; killing
```

### On the fixed, rebased tree

```
test db::tests::sub_cluster_by_tokens_drops_single_member_subclusters_after_split ... ok
test db::tests::a_threshold_that_cannot_advance_drops_instead_of_looping ... ok
test db::tests::test_sub_cluster_by_tokens ... ok
test db::tests::oversized_space_cluster_resplits_in_steps_instead_of_dropping ... ok
test db::tests::resplit_children_keep_their_base_group_position_for_max_clusters_ties ... ok
test result: ok. 5 passed; 0 failed
rc_clippy=0 (wenlan-core, lib + tests, no warnings)   rc_check=0 (wenlan-server)   cargo fmt clean
```

PR #597, opened 2026-08-25
fix(cli,core,mcp): say what to do when the daemon, model, or config is broken

7xuanlu:
## Why

Failure-mode sweep of the first hour against the released v0.17.0 binaries (report: `wenlan-failure-sweep-2026-08-25.md`, twelve scenarios in an isolated data root; the developer daemon was never contacted). Four common breakages left the user with no next step; two more were silent or misleading.

| Breakage | v0.17.0 showed | Now |
|---|---|---|
| Daemon down, `wenlan recall`/`search`/`brief` | `Error: GET …/api/spaces failed` + `Caused by: … Connection refused (os error 61)` | First line: *no Wenlan daemon is listening at `<url>` — run `wenlan status`; if Wenlan is installed, `wenlan background on` starts it* (notes when `WENLAN_NO_AUTOSTART` suppressed the automatic start; a remote `WENLAN_HOST` gets *cannot connect to the Wenlan daemon at `<url>` (from WENLAN_HOST)*). The reqwest chain stays underneath. |
| `WENLAN_HOST=17917` or `host:port` | `GET 17917/api/health failed (is the daemon running?)` | The shorthand is completed: digits → `http://127.0.0.1:<port>`, schemeless host → `http://<host>`, empty → default. |
| Daemon exited on purpose (newer schema, missing model, unwritable data root) | `wenlan doctor`: "Daemon: not reachable" + a paragraph about the newer-schema case regardless of cause | Doctor prints the last `ERROR` line of the daemon log (colour codes stripped), shows the schema paragraph only when that is the cause, otherwise says to fix the cause and run `wenlan background on`. |
| First start, no model cached, no network | `Embedding error: init embedder: Failed to retrieve model_optimized.onnx` | *could not load the embedding model (quantized BGE base en v1.5, about 210 MB) from `<cache>`. The first start downloads it from huggingface.co; check the network or proxy, then start Wenlan again. Cause: `<error>`* (cache dir follows `HF_HOME` like fastembed; a damaged cache, permission, or ONNX Runtime failure gets different advice) |
| `config.json` with a JSON typo | Silently replaced by defaults, nothing logged | `[config] <path> is not valid JSON (<error>); using defaults until it is fixed` |
| MCP tool call, daemon down | "Install the local Wenlan runtime…" (assumes not installed) | Leads with *If Wenlan is already installed, the daemon may just be stopped: run `wenlan status`, then `wenlan background on`.* |

Scenarios that were already clear and are unchanged: port taken by a healthy daemon (exit 75 for launchd), port taken by a mute holder, a store migrated by a newer Wenlan (the daemon's refusal text), an unwritable data root, capture while offline (outbox), a stale 0.16.0 CLI.

## Verification

Focused tests (`cargo test -p wenlan --lib -- origin_host_tests connect_failure doctor_tests recovery::tests client::tests`): 14 passed, including a live connect failure against `127.0.0.1:1` asserting the new first line. `cargo test -p wenlan-core --lib -- config::tests`: 30 passed. Clippy clean on `wenlan`, `wenlan-core`, `wenlan-mcp` (`--all-targets`); `cargo check -p wenlan-server` clean; `cargo fmt` clean.

Real-surface proof with the freshly built binaries is in the first comment.

Not run: autostart with no registered service (the recovery path talks to the real launchd session, which this sweep must not touch), disk full, and the desktop app.

### DIFF
--- a/crates/wenlan-cli/README.md
+++ b/crates/wenlan-cli/README.md
@@ -51,6 +51,9 @@ Set `WENLAN_HOST` to point at a remote daemon:
 export WENLAN_HOST=http://127.0.0.1:7878  # default
 ```
 
+A bare port (`17917`) or `host:port` is completed to `http://127.0.0.1:17917` /
+`http://host:port`; a loopback shorthand takes part in autostart like the full URL.
+
 When a request to a loopback daemon cannot connect, the CLI automatically starts
 the registered background service (`com.wenlan.server`) and waits for its health
 endpoint. Remote or invalid `WENLAN_HOST` values keep the original connection
--- a/crates/wenlan-cli/src/client.rs
+++ b/crates/wenlan-cli/src/client.rs
@@ -145,9 +145,20 @@ impl WenlanClient {
                 };
                 let original = anyhow::Error::new(error).context(what.to_owned());
                 if let Err(recovery_error) = recovery::recover(&self.base_url).await {
-                    return Err(original.context(format!("{recovery_error:#}")));
+                    return Err(original.context(format!("{recovery_error:#}")).context(
+                        "the daemon did not come up on its own — run `wenlan doctor` to see why",
+                    ));
                 }
-                retry.send().await.with_context(|| what.to_owned())
+                retry
+                    .send()
+                    .await
+                    .with_context(|| what.to_owned())
+                    .with_context(|| {
+                        format!(
+                            "the daemon was started but {} still does not answer — run `wenlan doctor`",
+                            self.base_url
+                        )
+                    })
             }
             Err(error) => Err(error).with_context(|| what.to_owned()),
         }
@@ -658,18 +669,23 @@ mod tests {
 
     #[tokio::test]
     async fn connect_failure_says_what_to_do_next() {
-        let client = client_for("http://127.0.0.1:1").with_recovery(false);
-        let error = client
-            .health()
+        // An ephemeral port that was just released: nothing listens there.
+        let listener = std::net::TcpListener::bind("127.0.0.1:0").expect("ephemeral port");
+        let addr = listener.local_addr().expect("local addr");
+        drop(listener);
+        let base = format!("http://{addr}");
+        let client = client_for(&base).with_recovery(false);
+        let error = tokio::time::timeout(std::time::Duration::from_secs(10), client.health())
             .await
-            .expect_err("nothing listens on port 1");
+            .expect("connect failure within 10s")
+            .expect_err("nothing listens on the released port");
         let text = format!("{error:#}");
         assert!(
-            text.contains("no Wenlan daemon is listening at http://127.0.0.1:1"),
+            text.contains(&format!("no Wenlan daemon is listening at {base}")),
             "{text}"
         );
         assert!(
-            text.contains("GET http://127.0.0.1:1/api/health failed"),
+            text.contains(&format!("GET {base}/api/health failed")),
             "{text}"
         );
     }
--- a/crates/wenlan-cli/src/client.rs
+++ b/crates/wenlan-cli/src/client.rs
@@ -139,7 +139,9 @@ impl WenlanClient {
             Ok(response) => Ok(response),
             Err(error) if error.is_connect() => {
                 let Some(retry) = retry else {
-                    return Err(error).with_context(|| what.to_owned());
+                    return Err(error)
+                        .with_context(|| what.to_owned())
+                        .with_context(|| recovery::connect_failure_hint(&self.base_url));
                 };
                 let original = anyhow::Error::new(error).context(what.to_owned());
                 if let Err(recovery_error) = recovery::recover(&self.base_url).await {
@@ -652,6 +654,24 @@ mod tests {
         let request = build_list_request(Some(20), None, None);
         let json = serde_json::to_value(request).expect("serialize list request");
         assert!(json.get("confirmed").is_none());
+    }
+
+    #[tokio::test]
+    async fn connect_failure_says_what_to_do_next() {
+        let client = client_for("http://127.0.0.1:1").with_recovery(false);
+        let error = client
+            .health()
+            .await
+            .expect_err("nothing listens on port 1");
+        let text = format!("{error:#}");
+        assert!(
+            text.contains("no Wenlan daemon is listening at http://127.0.0.1:1"),
+            "{text}"
+        );
+        assert!(
+            text.contains("GET http://127.0.0.1:1/api/health failed"),
+            "{text}"
+        );
     }
 
     fn client_for(base_url: &str) -> WenlanClient {
--- a/crates/wenlan-cli/src/client/lint.rs
+++ b/crates/wenlan-cli/src/client/lint.rs
@@ -18,7 +18,9 @@ pub fn origin_host_from_env() -> String {
 /// without a base", which the CLI used to report as "is the daemon
 /// running?". Complete the obvious shorthand instead: digits become a
 /// loopback URL, a schemeless host gets `http://`, and an empty value means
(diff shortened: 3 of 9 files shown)


7xuanlu:
## Real-surface proof (freshly built `target/debug/wenlan` and `wenlan-server`, same isolation as the sweep)

Daemon down, `wenlan recall hello` (`WENLAN_HOST=http://127.0.0.1:17999`, `WENLAN_NO_AUTOSTART=1`):

```
Error: no Wenlan daemon is listening at http://127.0.0.1:17999 — run `wenlan status`; if Wenlan is installed, `wenlan background on` starts it (WENLAN_NO_AUTOSTART is set, so it was not started automatically)

Caused by:
    0: GET http://127.0.0.1:17999/api/spaces failed
    1: error sending request for url (http://127.0.0.1:17999/api/spaces)
    2: client error (Connect)
    3: tcp connect error
    4: Connection refused (os error 61)
```

`WENLAN_HOST=17917 wenlan status` and `WENLAN_HOST=127.0.0.1:17917 wenlan status` (live isolated daemon on 17917): both now return `{"status": "ok", "db_initialized": true, "version": "0.17.0"}` (v0.17.0 answered `GET 17917/api/health failed (is the daemon running?)`).

`wenlan doctor` against the data root whose daemon refused a newer schema:

```
Daemon: not reachable on http://127.0.0.1:17926/api/health
  Last daemon error (…/fm/data-newer/logs/wenlan-server.log):
    2026-08-25T07:01:30.284818Z ERROR wenlan_server: wenlan-server terminated with an error: Vector DB error: database schema is version 9999, newer than this build supports (125); a newer Wenlan has already migrated it. Upgrade Wenlan (or quit the older copy) …
  If the store was migrated by a NEWER Wenlan, this build refuses to open it on purpose (this build supports schema version 125).
  Check the daemon log for: "newer than this build supports".
  Fix: upgrade Wenlan, or quit the newer copy. Restarting will not clear it.
```

…against the data root whose daemon could not load the model:

```
Daemon: not reachable on http://127.0.0.1:17927/api/health
  Last daemon error (…/fm/data-nomodel/logs/wenlan-server.log):
    2026-08-25T07:01:53.016315Z ERROR wenlan_server: wenlan-server terminated with an error: Embedding error: init embedder: Failed to retrieve model_optimized.onnx
  Fix the cause above, then run `wenlan background on` (or open the Wenlan app).
```

…and with no log at all:

```
Daemon: not reachable on http://127.0.0.1:17999/api/health
  No daemon error recorded in …/fm/data-empty/logs/wenlan-server.log.
  Start it with `wenlan background on` (or open the Wenlan app); if it stops again, that log says why.
```

New daemon, fresh data root, empty `HF_HOME`, network to `huggingface.co` denied by the sandbox (exit 1):

```
wenlan-server terminated with an error: Embedding error: could not load the embedding model (BGE base v1.5, about 140 MB): Failed to retrieve model_optimized.onnx: request error: https://us.aws.cdn.hf.co/xet-bridge-us/… The first start downloads it from huggingface.co; check the network or proxy, then start Wenlan again
```

(The `from <cache>` part is omitted here because a fresh store has no per-DB cache yet and fastembed used `HF_HOME`.)

New daemon with `config.json` = `{ "page_map_auto_suggest": tru` (daemon log):

```
WARN wenlan_core::config: [config] …/fm/data-badcfg2/config.json is not valid JSON (EOF while parsing a value at line 1 column 30); using defaults until it is fixed
```

7xuanlu:
## Review pass (one adversarial pass, Codex Sol, `xhigh`) — verdict MERGE, no P1

Five P2 and four P3 findings, all acted on in the follow-up commit.

| Finding | What changed |
|---|---|
| P2 doctor reads a log that does not exist on Linux/Windows, never reads the macOS fallback log, and resolves the data root without the legacy env names | `wenlan doctor` uses `wenlan_core::config::data_root()`; reads the daemon log on macOS only (primary root, then the fallback root when the primary log is absent); Linux/Windows are told the service console holds the log (`journalctl --user -u wenlan-server`) and that a store migrated by a newer Wenlan is refused on purpose. |
| P2 doctor presents stale errors as current | Checks the `autostart.off` marker first ("The background service is switched off (`wenlan background off`)…"); only the lines after the last `wenlan-server v…` banner count; an error followed by `graceful shutdown complete` is not reported. Regression test `an_error_from_a_previous_run_or_before_a_clean_shutdown_is_not_current`. |
| P2 embedder message: wrong size, wrong cache when `HF_HOME` is set, network advice for every failure | "quantized BGE base en v1.5, about 210 MB" (the file is 207.7 MiB on disk); the directory is `HF_HOME` when set, as fastembed does; advice comes before the raw cause; permission / damaged-cache / ONNX Runtime failures get their own advice. |
| P2 corrupt `config.json` warns every 30 s | Warned once per distinct (path, error); again only after the error changes or a good parse in between. |
| P2 autostart branches give no next step | "the daemon did not come up on its own — run `wenlan doctor` to see why" / "the daemon was started but `<url>` still does not answer — run `wenlan doctor`". `status --format json` now carries the whole chain in `error`. |
| P3 MCP hint assumes the CLI is on PATH | Names the app, a CLI on PATH, and a connector-only `wenlan-mcp` install separately. |
| P3 numeric `WENLAN_HOST` accepts `0` and enables autostart silently | Digits must parse as a non-zero `u16`; `0` / `70000` are left as typed. README documents the shorthand and that a loopback shorthand takes part in autostart. |
| P3 test on privileged port 1 without timeout | Binds `127.0.0.1:0`, releases it, connects under a 10 s `tokio::time::timeout`. |
| P3 one bad byte hides the log | `from_utf8_lossy`; a long line keeps head and tail so the advice at the end survives. |

Not done: reading the systemd journal from doctor, a Windows-specific diagnostic, and a metadata fingerprint for the config guard (the error text is enough to dedupe).

## Receipts

`cargo fmt` clean; `cargo clippy --all-targets` clean on `wenlan` and `wenlan-mcp`, `cargo clippy -p wenlan-core` clean; `cargo test -p wenlan --lib -- origin_host_tests connect_failure doctor_tests recovery::tests client::tests`: 15 passed; `cargo test -p wenlan-core --lib -- config::tests`: 30 passed; `cargo check -p wenlan-server` clean.

Real surface, rebuilt binaries, inside the sandbox (network to `huggingface.co` denied), fresh data root and empty `HF_HOME`:

```
wenlan-server terminated with an error: Embedding error: could not load the embedding model (quantized BGE base en v1.5, about 210 MB) from /…/fm/hf-empty3. The first start downloads it from huggingface.co; check the network or proxy, then start Wenlan again. Cause: Failed to retrieve model_optimized.onnx: request error: https://us.aws.cdn.hf.co/xet-bridge-us/…
```

`wenlan doctor` on that data root:

```
Daemon: not reachable on http://127.0.0.1:17932/api/health
  Last daemon error (/…/fm/data-nomodel3/logs/wenlan-server.log):
    2026-08-25T07:59:44.258532Z ERROR wenlan_server: wenlan-server terminated with an error: Embedding error: could not load the embedding model (quantized BGE base en v1.5, about 210 MB) from /…/fm/hf-empty3. The first start downloads it from huggingface.co; check the n … Iz7DYFNBkQpPhg__&Key-Pair-Id=01KXEF4KZ1B6FV465MAWR4M21F: Proxy failed to connect
  Fix the cause above, then run `wenlan background on` (or open the Wenlan app).
```

Corrupt `config.json`, daemon start: the `[config] … is not valid JSON (EOF while parsing a value at line 1 column 30); using defaults until it is fixed` warning appears once (it appeared twice before the guard). The other proofs from the first comment (daemon-down hint, `WENLAN_HOST=17917`, schema-refusal doctor, fallback log when the data root was unwritable) reproduce unchanged with the rebuilt binaries.

PR #602, opened 2026-08-26
fix(release): ship wenlan-server in the Homebrew wenlan formula

7xuanlu:
Pre-launch tracker row 12 (`wenlan-audit-2026-08-22.md`), first-run gauntlet finding F7.

## What was wrong

`brew install 7xuanlu/tap/wenlan` installed from `wenlan-cli-darwin-arm64.tar.gz`, which holds only the `wenlan` binary (39 MB). The documented next step, `wenlan background on`, looks for `wenlan-server` beside the CLI (`current_server_path` in `crates/wenlan-cli/src/commands/service.rs`) and failed with `wenlan-server not found next to origin at /opt/homebrew/bin/wenlan-server. Re-run the Wenlan installer.` The gauntlet's `brew-background-on` row was red on v0.17.0. The full `wenlan-darwin-arm64.tar.gz` archive already carries `wenlan`, `wenlan-server` (87 MB), and `wenlan-mcp`.

## What changed

| File | Change |
|---|---|
| `.github/workflows/release.yml` | The `homebrew-artifacts` upload carries `wenlan-darwin-arm64.tar.gz` instead of the CLI-only archive; `update-homebrew` hashes it as `SHA_DARWIN_ARM64`; the `wenlan` formula installs from it with `bin.install "wenlan", "wenlan-server"` and its test also runs `wenlan-server --help`. `wenlan-mcp` keeps its own formula, so installing both stays conflict-free. Both formula descriptions get a real em dash instead of the double-encoded `â€”` that has been shipping in `brew info`. |
| `scripts/release-workflow-contract.test.py` | New markers: the promote job must upload the full archive; the Homebrew job must hash it, install both binaries, and test the server; the CLI-only archive name may not reappear in the Homebrew job. |
| `scripts/first-run/brew.sh` | New `brew-server-installed` row; comments no longer call the `background on` failure expected; the status row after `background on` is renamed. The CLI and MCP loops keep using the scratch daemon from the archive. |
| `crates/wenlan-cli/src/commands/service.rs` | The missing-daemon error says `next to wenlan` and names the fix per install path: `brew upgrade 7xuanlu/tap/wenlan`, otherwise `npx wenlan setup`. |

Not changed: the two darwin archives, `release_targets`, and the asset validator; `wenlan-cli-darwin-arm64.tar.gz` is still published for direct downloads.

## Receipts

- `python3 scripts/release-workflow-contract.test.py`: `PASS: release promotion, Homebrew, and Node 24 action contracts`.
- Mutation control: with the formula edited back to `bin.install "wenlan"`, the same test reports `Homebrew wenlan formula omits 'bin.install "wenlan", "wenlan-server"'`; restored, PASS again.
- `python3 scripts/release_targets.test.py`: OK.
- Rendered `wenlan.rb` (placeholders substituted the way the job does it): `ruby -c` → `Syntax OK`.
- `shellcheck scripts/first-run/brew.sh`: only the pre-existing SC1091 info about sourcing `lib.sh`.
- `cargo check -p wenlan` clean.
- `target/debug/wenlan-server --help` prints `Usage: wenlan-server [OPTIONS]`, so the new formula assertion matches.

End-to-end proof is the formula the next release generates, then the gauntlet's brew rows against that release (tracker row 22).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -885,7 +885,7 @@ jobs:
         with:
           name: homebrew-artifacts
           path: |
-            ${{ runner.temp }}/promoted-assets/wenlan-cli-darwin-arm64.tar.gz
+            ${{ runner.temp }}/promoted-assets/wenlan-darwin-arm64.tar.gz
             ${{ runner.temp }}/promoted-assets/wenlan-mcp-darwin-arm64.tar.gz
           retention-days: 1
           if-no-files-found: error
@@ -1499,14 +1499,14 @@ jobs:
           require_event_tag
           VERSION="${RELEASE_TAG#v}"
           SHA_MCP_DARWIN_ARM64=$(shasum -a 256 wenlan-mcp-darwin-arm64.tar.gz | cut -d' ' -f1)
-          SHA_CLI_DARWIN_ARM64=$(shasum -a 256 wenlan-cli-darwin-arm64.tar.gz | cut -d' ' -f1)
+          SHA_DARWIN_ARM64=$(shasum -a 256 wenlan-darwin-arm64.tar.gz | cut -d' ' -f1)
 
           git clone "https://x-access-token:${HOMEBREW_TAP_TOKEN}@github.com/7xuanlu/homebrew-tap.git" tap
           mkdir -p tap/Formula
 
           cat > tap/Formula/wenlan-mcp.rb << 'FORMULA'
           class WenlanMcp < Formula
-            desc "MCP server for Wenlan â€” personal agent memory layer"
+            desc "MCP server for Wenlan — personal agent memory layer"
             homepage "https://github.com/7xuanlu/wenlan"
             version "VERSION_PLACEHOLDER"
             url "https://github.com/7xuanlu/wenlan/releases/download/vVERSION_PLACEHOLDER/wenlan-mcp-darwin-arm64.tar.gz"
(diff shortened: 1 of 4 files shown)


7xuanlu:
## Review follow-up (31fdc9f1)

One adversarial Codex Sol pass on 2605f1c7 returned MERGE with two smaller findings. Both acted on:

- **Contract test scoped to the formula heredoc.** `heredoc_body` extracts the `tap/Formula/wenlan.rb` heredoc from the `update-homebrew` job; the assertions read its active (non-comment) lines for the full-archive `url`, `bin.install "wenlan", "wenlan-server"` and the `wenlan-server --help` test, forbid `wenlan-cli-darwin-arm64.tar.gz` inside the heredoc only, and check that the stamped sha256 comes from `SHA_DARWIN_ARM64`. Two committed mutations prove each check bites on its own: the install line moved into a comment, and only the URL pointed back at the CLI-only archive.
- **Teardown race.** The gauntlet's Homebrew teardown now waits up to 15 s for `wenlan-server` to exit after `launchctl bootout` before it uninstalls or deletes state.

Not done, on purpose: the Homebrew check still uses the real per-user data root (that is the documented first step it exercises) and the teardown still removes it, the same as the desktop-app check. The gauntlet is written for throwaway runners; making it non-destructive on a developer machine is a separate change.

Receipts on 31fdc9f1: `python3 scripts/release-workflow-contract.test.py` PASS (including the two new mutations), `shellcheck scripts/first-run/brew.sh` clean apart from the pre-existing "not following lib.sh" note.

PR #603, opened 2026-08-26
fix(app): let a newer Wenlan.app take over from a running older one

7xuanlu:
Pre-launch tracker row 13 (`wenlan-audit-2026-08-22.md`), first-run gauntlet findings F2 and F11.

## What was wrong

`tauri-plugin-single-instance` keys its socket by bundle id only, so opening a freshly installed 0.17.0 bundle while 0.16.0 ran made the newcomer exit after ~1 s and brought the old window to the front. Nothing told the user the old version was still what they were looking at; the README's upgrade story ("re-run the installer or open the new app") was false.

## What changed

| File | Change |
|---|---|
| `app/src/handover.rs` (new) | `newer_bundle_from_launch(argv, running)` reads the newcomer's `Info.plist` (argv[0] of a LaunchServices launch is `<bundle>.app/Contents/MacOS/<exe>`) and returns it only when `CFBundleShortVersionString` is strictly newer; anything else — a dev binary, missing or unparsable plist, same or older version — keeps the focus behavior. `relaunch_after_exit(bundle)` spawns a detached `/bin/sh` that waits for this pid to disappear, then `open`s the bundle: the old app cannot `open` the newcomer itself (LaunchServices would only activate it, and its socket would send the newcomer back to exit). |
| `app/src/lib.rs` | The single-instance callback checks for a newer bundle first (macOS only), records it, and runs the same guarded quit as the tray's "Quit Wenlan" (`request_full_quit`, falling back to `force_full_quit`). |
| `app/src/lifecycle.rs` | `set_handover_bundle`; `quit_origin` waits up to 5 s for the daemon to stop answering when a handover is pending (the new app starts its own daemon on the same port), then `exit_after_quit` schedules the reopen right before `app.exit(0)` on both exit paths. |
| `app/Cargo.toml`, `Cargo.lock` | `semver = "1"` (already in the tree through Tauri) for the version comparison. |
| `scripts/install-macos-app.sh` | Apps up to 0.17.0 have no handover, so the installer asks a Wenlan still running from the replaced bundle to quit (`osascript … quit`), waits up to 10 s, and warns if it is still there before `open`. |
| `README.md` | Says how to upgrade by drag and drop (and that 0.17.0 and older must be quit by hand), and names the Gatekeeper dialog a first launch shows before pointing at "Open Anyway" (F11 — the notarization itself stays a known gap). |

Why a full quit rather than a plain exit: "Quit Wenlan" already shuts the daemon down and unloads the LaunchAgents, and the new app's startup preflight installs them again — so the handover is the manual "quit, then open the new one" made automatic, on the app's most-exercised exit path.

## Receipts

- `cargo test -p wenlan-app --lib`: 438 passed, 0 failed, 1 ignored (3 new `handover` tests: newer/equal/older bundle, non-bundle argv and missing or bad plist, relaunch script shape).
- `cargo clippy -p wenlan-app --all-targets -- -D warnings` clean; `cargo fmt --check -p wenlan-app` clean.
- `bash scripts/test-install-macos-app.sh`: `install-macos-app tests passed`; `shellcheck scripts/install-macos-app.sh` clean.

Unchecked: the live handover between two signed bundles. It needs two built versions on a Mac; the next release's gauntlet upgrade walk (tracker row 22) is where that runs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9070,6 +9070,7 @@ dependencies = [
  "raw-window-handle",
  "regex",
  "reqwest 0.12.28",
+ "semver",
  "serde",
  "serde_json",
  "serial_test",
--- a/README.es-ES.md
+++ b/README.es-ES.md
@@ -1,4 +1,4 @@
-<!-- README_SYNC: source=README.md sha256=5e3c459d2b63607afbca51d7c7c6695326138156a9be535bcd02ef675c37a927 -->
+<!-- README_SYNC: source=README.md sha256=c8dbe1b39a702cde2a79dfc1b7f8b7c836fa3ee3efe5a6ac6e77d084ca31e413 -->
 
 <p align="center">
   <picture>
@@ -55,7 +55,7 @@ Wenlan funciona como un único daemon local. La aplicación de escritorio lo lle
 
 No hay nada más que instalar. La aplicación incluye el daemon, la CLI y el conector MCP, arranca el daemon al abrirse y ofrece conectar los clientes de IA que detecta: el plugin para Claude Code y Codex, una entrada MCP para el resto. A partir de ahí lees Páginas, inspeccionas la Fuente detrás de cualquier cita y gestionas el sistema de conocimiento.
 
-Esta vista previa aún no está notarizada, así que macOS bloquea el primer arranque con "Apple no ha podido verificar que Wenlan no contenga software malicioso". Pulsa Listo y luego permítelo una vez en Ajustes del Sistema, Privacidad y seguridad, "Abrir de todos modos". Un comando se salta ese paso: comprueba la descarga contra el SHA-256 publicado en GitHub, elimina la cuarentena solo para esta aplicación y no cambia ninguna configuración de seguridad de macOS.
+Esta vista previa está firmada ad hoc y no notarizada, así que macOS bloquea el primer arranque: puede decir que Wenlan "no se puede abrir" o que no ha podido comprobar si contiene software malicioso. Cierra ese aviso y luego permite la aplicación una vez en Ajustes del Sistema, Privacidad y seguridad, "Abrir de todos modos". Un comando se salta ese paso: comprueba la descarga contra el SHA-256 publicado en GitHub, elimina la cuarentena solo para esta aplicación y no cambia ninguna configuración de seguridad de macOS.
 
 ```bash
 /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/7xuanlu/wenlan/main/scripts/install-macos-app.sh)"
--- a/README.es-ES.md
+++ b/README.es-ES.md
@@ -1,4 +1,4 @@
-<!-- README_SYNC: source=README.md sha256=c4e73e02e9e5607b16b7e0d2282e1bc5cc3aff927b0e5acb7bbdfb038a688fc5 -->
+<!-- README_SYNC: source=README.md sha256=5e3c459d2b63607afbca51d7c7c6695326138156a9be535bcd02ef675c37a927 -->
 
 <p align="center">
   <picture>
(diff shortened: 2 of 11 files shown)


7xuanlu:
Review follow-up, pushed as `2b6ca5ca`.

**Acted on**

- Handover deadline: after the guarded quit request succeeds, `arm_handover_deadline` waits 15 s (`handover::QUIT_DEADLINE`) and forces the quit if a handover is still pending, no quit is running, and the guard is still in flight (`force_if_in_flight`). A frontend that acknowledged and never called the full quit can no longer leave the old app hidden. Test: `an_acknowledged_guard_that_never_resolves_is_still_in_flight_for_the_handover_deadline`.
- Refused quit: `cancel_guarded_quit_request` now clears the pending handover (`lifecycle::clear_handover_bundle`), so the deadline leaves it alone and a later quit does not reopen the newer bundle.
- `exit(1)` bypass: `exit_after_quit(app, code)` is the single exit path; `force_full_quit`'s failure branch goes through it with code 1, so the reopen still happens.
- Daemon wait: `wait_for_daemon_to_stop` wraps the whole wait in `tokio::time::timeout(limit)` (the health calls are inside the ceiling), then probes `host:port` with `TcpStream::connect` until the listener is closed. Test: `listener_addr_is_the_authority_of_the_base_url`.
- Build metadata: `newer_bundle_from_launch` uses `cmp_precedence`. Test: `build_metadata_never_decides_an_upgrade`.
- Installer: the quit now happens before the bundle is replaced; it matches any running bundle (`pgrep -f '\.app/Contents/MacOS/wenlan-app$'`), runs `osascript` in the background and kills it after the 10 s wait, and `die`s ("the running Wenlan did not quit within 10 s...") instead of printing "Installed" when the app is still there. Driven with stubbed `pgrep`/`osascript`: not running, quits after three polls, Apple event hangs but the app quits (no stray osascript left), denied and never quits (rc=1 in 13 s), hangs and never quits (rc=1, osascript killed).
- README: "ad-hoc signed and not notarized", both Gatekeeper messages named; the three translations updated and their sync markers restamped (`check-readme-translations.py` rc=0).

Receipts: `cargo fmt --check -p wenlan-app` rc=0, `cargo clippy -p wenlan-app --all-targets -- -D warnings` rc=0, `cargo test -p wenlan-app --lib` 441 passed / 0 failed / 1 ignored.

**Not done here, on purpose**

- Isolated handover leaving the app-owned sidecar alive: fixed by the next PR (sidecar lifetime, branch `fix/sidecar-lifetime`, stacked on this one), which stops the sidecar on every exit path including that branch.
- The in-app updater's `app.restart()` keeping the old launchd daemon is pre-existing and outside this change; noted for a follow-up.
- A live LaunchServices handover between two built bundles was not run; the path stays `unchecked` beyond the unit tests above.
- The installer script has no test harness in the repo today; the stubbed drive above is a receipt, not a durable test.

7xuanlu:
Correction to the comment above: the installer does have a test harness in the repo, `scripts/test-install-macos-app.sh`, and CI runs it (`ci.yml`, the app-check path filter). I ran it against this branch's installer before this commit: the three existing cases pass.

The new commit adds two cases for the quit path with `pgrep` and `osascript` stubbed on `PATH`, the way the interrupted-install case stubs `mv`:

- a running Wenlan that quits when asked is replaced (the installer sent `tell application id "com.wenlan.desktop" to quit`, the old bundle is gone);
- a Wenlan that never quits fails the install after 10 s with `did not quit within 10 s`, and the old app is untouched.

`bash scripts/test-install-macos-app.sh`: 5 cases, `install-macos-app tests passed`. Mutation control: the same test file against the pre-PR installer (`43c00470`) fails on the quit case. `shellcheck -S warning` clean.

PR #604, opened 2026-08-26
fix(app): stop the sidecar the app spawned and install outside the data root on Windows

7xuanlu:
## What

The `wenlan-server` sidecar this app spawns is now the app's to stop, and the Windows installer no longer installs into the daemon's data root. Stacked on #603 (base is its branch until it merges).

First-run gauntlet findings: the sidecar outlived the app on macOS quit and Windows kill, and the NSIS default install dir collided with `%LOCALAPPDATA%\Wenlan`.

## How

- `app/src/daemon_start.rs`: the spawned `CommandChild` is remembered in a process-wide slot; `stop_sidecar()` takes it and, on Unix, sends SIGTERM and polls up to 3 s before killing; on Windows it kills outright. The child is forgotten when its `Terminated` event arrives. On Windows the child is also bound to a kill-on-close Job Object (`windows-sys 0.59`), so even a hard-killed app takes the daemon with it.
- `app/src/lib.rs`: a SIGTERM handler (Unix) stops the sidecar and exits 0 without touching LaunchAgents.
- `app/src/lifecycle.rs`: `quit_origin` stops the sidecar before exiting, both in the isolated run and on the normal path (covers the "isolated handover leaves the sidecar alive" finding on #603).
- `app/windows/installer-hooks.nsh` + `app/tauri.windows.conf.json`: `NSIS_HOOK_PREINSTALL` moves the default `$INSTDIR` from `$LOCALAPPDATA\Wenlan` to `$LOCALAPPDATA\Programs\Wenlan`. An explicit user choice is kept.
- `scripts/first-run/windows-nsis.ps1`: looks for the app under `Programs\Wenlan` and asserts the install dir is outside the data root.
- `docs/cross-platform.md`: the install-dir and sidecar-lifetime notes.

## Verification

- `cargo fmt --check -p wenlan-app` rc=0; `cargo clippy -p wenlan-app --all-targets -- -D warnings` rc=0; `cargo test -p wenlan-app --lib` 441 passed / 0 failed / 1 ignored.
- Windows code: the `job` module compiled and clippy-clean with `--target x86_64-pc-windows-msvc` in a scratch crate; a mutation control (wrong pid type) failed as expected.
- Live, isolated macOS debug app (`WENLAN_PORT=17993`, scratch `HOME`/`WENLAN_DATA_DIR`, `WENLAN_NO_AUTOSTART=1`): after SIGTERM the log shows `[app] SIGTERM received; stopping the sidecar and exiting` and `sidecar 66675 exited after SIGTERM`; the app and sidecar pids were gone and port 17993 closed. The guarded-quit route (tray/Cmd-Q) calls the same `stop_sidecar` but was not driven live: an Apple event cannot target the unbundled debug binary. `unchecked` live for that route.
- The NSIS hook and the Windows first-run check run only in the release `app-bundle-windows` job and the gauntlet; `unchecked` here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9097,6 +9097,7 @@ dependencies = [
  "tracing-subscriber",
  "uuid",
  "wenlan-types",
+ "windows-sys 0.59.0",
  "zip 2.4.2",
 ]
 
--- a/app/Cargo.toml
+++ b/app/Cargo.toml
@@ -81,8 +81,8 @@ sysinfo = { workspace = true }
 pdf-extract = "0.10"
 zip = { version = "2.2", default-features = false, features = ["deflate"] }
 
-# Keep this table LAST. A TOML table runs until the next header, so any
-# dependency written below it silently becomes macOS-only. That is how
+# Keep the target tables LAST. A TOML table runs until the next header, so any
+# dependency written below one silently becomes platform-specific. That is how
 # wenlan-types, sysinfo, pdf-extract, and zip stopped existing off macOS and
 # left the app crate uncompilable on Windows (15 unresolved-import errors,
 # unnoticed because app-check only runs on macOS).
@@ -90,6 +90,10 @@ zip = { version = "2.2", default-features = false, features = ["deflate"] }
 cocoa = "0.26"
 objc = "0.2"
 raw-window-handle = "0.6"
+
+# Job object that ends the sidecar daemon with the app (daemon_start::job).
+[target.'cfg(windows)'.dependencies]
+windows-sys = { version = "0.59", features = ["Win32_Foundation", "Win32_Security", "Win32_System_JobObjects", "Win32_System_Threading"] }
 
 [dev-dependencies]
 serial_test = "3"
--- a/app/src/daemon_start.rs
+++ b/app/src/daemon_start.rs
@@ -14,8 +14,9 @@
 //! and [`stop_sidecar`] ends it on quit, and on Windows the process is bound
 //! to a kill-on-close job object so a hard kill of the app takes it too.
 
-use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
 use std::sync::{Arc, Mutex};
+use std::time::Duration;
 use tauri_plugin_shell::process::CommandChild;
 use tokio::sync::RwLock;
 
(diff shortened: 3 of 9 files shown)


7xuanlu:
Adversarial review (one pass, Codex Sol): DO NOT MERGE on one P1, now fixed; the rest acted on or skipped as listed.

- P1, fixed: an in-app update of an existing Windows install relocated the new files while the old copy and its shortcuts stayed behind (Tauri's `/UPDATE` run skips the old uninstaller and keeps shortcuts). `app/windows/installer-hooks.nsh` now moves `$INSTDIR` only when `$UpdateMode <> 1`; an install that predates the move stays put until the user runs the full installer. `docs/cross-platform.md` says so.
- P2 (Unix), fixed: `stop_sidecar` no longer sends a raw SIGTERM to the numeric PID (the shell plugin's wait thread can reap the child first, and the PID can be reused). It posts `/api/shutdown`, waits up to 3 s for the port to close (`wait_for_daemon_to_stop`, now shared and platform-free), and only then kills through the `CommandChild` handle, which is a no-op once the child was reaped.
- P3, fixed: the slot carries a spawn generation, a late `Terminated` event cannot clear a newer child, and a second spawn kills the child still in the slot. A second SIGTERM exits immediately; a stop that exceeds 10 s exits anyway.
- P3, fixed: the docs no longer claim the daemon never outlives the app; service-owned daemons and a failed job-object binding are named.
- P2 (Windows job assignment fails open), skipped: refusing to run the daemon at all on a machine whose parent job forbids nesting would cost the user their first hour outright; the failure is logged and the quit/SIGTERM path still ends the sidecar through the handle. Only an app crash on such a machine is uncovered, as documented.
- P3 (gauntlet script resolving the registry `InstallLocation` first), skipped: cannot be run here, and the P1 gate removes the two-copy case it was meant to catch.
- The two-version updater test needs a Windows runner: `unchecked`.

Receipts: `cargo test -p wenlan-app --lib` 441 passed; `cargo clippy -p wenlan-app --all-targets -- -D warnings` and `cargo fmt --check` clean. Live, isolated debug app on macOS (scratch port, data dir, `HOME`): `kill -TERM` on the app logged `[daemon-start] sidecar 7712 released the port after the shutdown request`, the app exited within 1 s, the sidecar exited, the port closed; the real daemon on 7878 was untouched. The Windows build stays `unchecked` locally (no target installed); the `platform-compile` CI job is the gate.

PR #605, opened 2026-08-26
fix(plugin): warn on daemon drift by release part and give a first boot 60 s

7xuanlu:
## What

The plugin's SessionStart hook never warned about daemon version drift on a real install, and the setup skill's re-probe after bootstrap gave up before a first boot could finish. First-run gauntlet findings on plugin drift and the re-probe.

## Why

- `plugin/hooks/check-daemon.sh` treated any `+g<sha>` build-metadata suffix as a dev build and exited silently. Every published daemon carries that suffix: the binaries are built on the release PR, before the `v*` tag exists, so `head_on_release_tag()` in `crates/wenlan-core/build.rs` is false at build time. (Filed separately as a release-stamping follow-up; this PR makes the consumers ignore metadata, which is what semver says to do.)
- `plugin/skills/setup/SKILL.md` step 4 polled for five seconds; a first boot downloads the embedding model first.

## How

- Hook: compare release parts (`${ver%%+*}`), then major.minor exactly as before. The exemption is gone.
- `plugin/hooks/test-check-daemon.sh`: a loopback `python3 -m http.server` answers `/api/health` with a chosen version; four drift cases (metadata + minor drift warns, metadata alone is quiet, patch drift is quiet, a source build a minor behind warns).
- SKILL.md step 4: poll up to 60 s, then print `wenlan doctor` output and stop; the causes list names the model download.
- `scripts/first-run/plugin-setup.sh`: the gauntlet check is `skill-reprobe-60s`, the loop as the skill writes it.
- `app/src/lib.rs`: the startup "Daemon version mismatch" warning compares the release part too (it fired on every launch against a published daemon).

## Verification

- `bash plugin/hooks/test-check-daemon.sh`: 8 passed, 0 failed. Mutation control: restoring the `*+g*) exit 0` exemption fails exactly the two "warns" cases (6 passed, 2 failed); restored, 8 passed. `shellcheck -S warning` clean on both hook scripts.
- `cargo fmt --check -p wenlan-app` rc=0; `cargo clippy -p wenlan-app --all-targets -- -D warnings` rc=0; `cargo test -p wenlan-app --lib` 436 passed / 0 failed / 1 ignored (includes `daemon_build_metadata_is_not_a_version_mismatch`).
- The 60 s skill loop and the gauntlet check run only in the first-run gauntlet workflow; `unchecked` here beyond `bash -n`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2296,6 +2296,8 @@ jobs:
           python3 scripts/validate-codex-plugin-slice.py
           python3 scripts/validate-plugin-contract.py
           bash scripts/validate-plugin-contract.test.sh
+      - name: Run the SessionStart hook tests
+        run: bash plugin/hooks/test-check-daemon.sh
       - name: Validate skill frontmatter
         run: |
           for f in plugin/skills/*/SKILL.md; do
--- a/app/src/lib.rs
+++ b/app/src/lib.rs
@@ -469,6 +469,14 @@ fn request_full_quit(app: &tauri::AppHandle) -> Result<(), tauri::Error> {
     Ok(())
 }
 
+/// The release part of a daemon version. Build metadata (`0.17.0+g1234abcd`)
+/// never counts as a mismatch: a source build carries it, and so can a
+/// published daemon whose binary was built before its tag existed.
+#[cfg(not(feature = "review-fixtures"))]
+fn release_part(version: &str) -> &str {
+    version.split('+').next().unwrap_or(version)
+}
+
 #[cfg(target_os = "macos")]
 fn startup_reveal_fallback_delay() -> std::time::Duration {
     std::time::Duration::from_millis(1200)
@@ -1072,7 +1080,7 @@ pub fn run() {
                             // path (LaunchAgent, sidecar, or dev checkout) —
                             // a stale one can hold the port and answer health
                             // while breaking newer API calls.
-                            if health.version != env!("CARGO_PKG_VERSION") {
+                            if release_part(&health.version) != env!("CARGO_PKG_VERSION") {
                                 log::warn!(
                                     "[init] Daemon version mismatch: daemon v{}, app v{} at {}; restart it (e.g. `wenlan restart`)",
                                     health.version,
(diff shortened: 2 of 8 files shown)


7xuanlu:
CI failed on `wenlan-types::plugin_distribution::plugin_setup_repairs_stale_daemon_versions`: the contract test reads `for i in 1 2 3` from the SessionStart hook, and the shellcheck-driven rename to `_` broke it. The follow-up commit restores the variable (warning silenced on that line) and, while here, gives `plugin-codex/skills/setup/SKILL.md` the same 60 s re-probe and doctor output as the Claude skill, so the two plugins do not drift.

`cargo test -p wenlan-types --test plugin_distribution`: 4 passed. `bash plugin/hooks/test-check-daemon.sh`: 8 passed. `shellcheck -S warning` clean.

7xuanlu:
Adversarial review (one pass, Codex Sol): MERGE, with findings acted on in the commit above.

- MCP handshake kept the same `+g` skip this PR removes from the hook: `crates/wenlan-mcp/src/client.rs` now compares through semver in `handshake_message`; test `a_published_daemon_with_build_metadata_still_gets_the_handshake` covers an older `0.16.0+g…` daemon, an equal one, and a newer one.
- The hook test was not a CI gate: `.github/workflows/ci.yml` plugin job runs `bash plugin/hooks/test-check-daemon.sh`. The stub server binds port 0 itself and publishes the port; readiness is asserted; the trap kills and waits.
- 60 s did not cover the documented first-boot budget: both setup skills poll for 240 s by wall clock, stop when `wenlan background on` already failed, and return non-zero after the doctor output; `scripts/first-run/plugin-setup.sh` check is `skill-reprobe-240s`.
- Hook guidance when the daemon is newer: "Run /wenlan:setup to repair: it updates an older runtime, or says when the plugin cache is stale, then verifies MCP." (the contract test's needle is kept).
- App compare strips build metadata on both sides.

Receipts: `bash plugin/hooks/test-check-daemon.sh` 8 passed; `cargo test -p wenlan-types --test plugin_distribution` 4 passed; `cargo test -p wenlan-mcp --lib` green including the new handshake test; `cargo test -p wenlan-app --lib version_mismatch` 1 passed; `cargo clippy -D warnings` clean on both crates; `shellcheck -S warning` clean. `actionlint` reports only pre-existing findings at other lines of ci.yml.

PR #607, opened 2026-08-26
fix(install): stop depending on anonymous GitHub API calls

7xuanlu:
## What

The install paths depended on anonymous GitHub API calls, which GitHub limits to 60 per hour per IP address. First-run gauntlet finding on the rate limit: `npx -y wenlan setup` and the plugin bootstrap's `install.sh` both failed with HTTP 403 before downloading anything.

## Why

A user behind a shared network (office, campus, CI) can hit the same 403 on their very first command, and nothing honored `GITHUB_TOKEN` or said what the 403 was.

## How

- `install.sh` and `crates/wenlan-cli/npm/run.js` make no API call at all: a pinned tag (`WENLAN_RELEASE_TAG`) is used as given, and the latest release comes from the redirect of `https://github.com/7xuanlu/wenlan/releases/latest` (a plain page, no API quota). The redirect is accepted only from `https://github.com/7xuanlu/wenlan/releases/tag/` on a 3xx, and every tag, pinned or resolved, must match `vX.Y.Z[-pre]`. A wrong tag now fails at the download with the release page named.
- `scripts/install-macos-app.sh` still needs the API for the asset's SHA-256 digest. It sends `Authorization: Bearer $GH_TOKEN` (or `GITHUB_TOKEN`) when set, and only to `https://api.github.com/`; on HTTP 403/429 it says it is GitHub's rate limit (wait for the reset or set `GITHUB_TOKEN`), and on 401 that the token is invalid or expired. The gauntlet's macOS app job passes `github.token`.
- `crates/wenlan-mcp/src/self_update_check.rs` reads the same redirect (no redirect following, `Location` header parsed) instead of the API. It also still pointed at `7xuanlu/origin`, the repository's old name.

## Verification

- `bash scripts/test-install-macos-app.sh`: 5 cases pass. A loopback server answering 403 proves the rate-limit message and that the token is withheld from a non-GitHub host; a recording fake `curl` proves `GH_TOKEN` reaches `api.github.com` as `Authorization: Bearer test-token` and that HTTP 401 names an expired token. The harness now runs in the `app-check` CI job. Mutation controls: the `main` installer fails with `installer did not name rate limiting`; sending the token to any `http*` URL fails with `installer sent the token to a host that is not api.github.com`.
- Live, in a scratch `HOME`: `install.sh` unpinned resolves `Latest release: v0.17.0` and installs; pinned `v0.17.0` installs without any lookup; pinned `v9.9.9` fails with the download URL and the release page named. `node run.js --version` unpinned downloads v0.17.0 and prints `wenlan 0.17.0`; pinned `v9.9.9` fails with the release page named.
- `cargo test -p wenlan-mcp --lib self_update_check`: 4 passed (includes the redirect parsing test); `cargo fmt --check` and `cargo clippy -p wenlan-mcp --all-targets -- -D warnings` clean. The live reqwest path of the MCP check is `unchecked` here; the redirect itself was confirmed with `curl -I` (302, `Location: …/releases/tag/v0.17.0`).
- `shellcheck -S warning` clean on the three shell files; `node --check` and `actionlint` clean.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -161,6 +161,10 @@ jobs:
               - 'playwright.review.config.ts'
               - 'scripts/prepare-sidecars.sh'
               - 'scripts/prepare-tauri-build-sidecars.sh'
+              # The macOS app installer and its harness need a macOS runner;
+              # app-check is the only job that runs them.
+              - 'scripts/install-macos-app.sh'
+              - 'scripts/test-install-macos-app.sh'
               # Deliberate pair: fnmatch vs picomatch zero-dir ** behavior.
               - 'scripts/*.test.ts'
               - 'scripts/**/*.test.ts'
@@ -2372,6 +2376,8 @@ jobs:
         run: cargo clippy -p wenlan-app --all-targets -- -D warnings
       - name: Rust tests (app)
         run: cargo test -p wenlan-app --lib
+      - name: macOS app installer tests
+        run: bash scripts/test-install-macos-app.sh
       - name: TypeScript check
         run: pnpm exec tsc -b
       - name: i18n gates
--- a/.github/workflows/first-run-gauntlet.yml
+++ b/.github/workflows/first-run-gauntlet.yml
@@ -337,6 +337,9 @@ jobs:
       GAUNTLET_OUT: ${{ github.workspace }}/gauntlet-out
       GAUNTLET_CHANNEL: macos-app
       REPO_ROOT: ${{ github.workspace }}
+      # The app installer's release lookup is a GitHub API call; the workflow
+      # token keeps repeated runs off the anonymous per-IP limit.
+      GITHUB_TOKEN: ${{ github.token }}
     steps:
       - uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6
       - name: Port 7878 precheck
--- a/crates/wenlan-cli/npm/run.js
+++ b/crates/wenlan-cli/npm/run.js
@@ -28,11 +28,29 @@ function binDir() {
   return path.join(os.homedir(), ".wenlan", "bin");
 }
 
-function githubApiUrl() {
-  if (REQUESTED_TAG) {
-    return `https://api.github.com/repos/${REPO}/releases/tags/${REQUESTED_TAG}`;
-  }
-  return `https://api.github.com/repos/${REPO}/releases/latest`;
+// The public releases page redirects to the latest tag. Unlike the GitHub API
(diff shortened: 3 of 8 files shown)


7xuanlu:
Adversarial review (one pass, Codex Sol): MERGE, with two P2 and three P3 findings, acted on as listed.

- P2, fixed: the bearer token went to any `http*` metadata URL, including plaintext and `WENLAN_APP_RELEASE_JSON_URL` overrides. `scripts/install-macos-app.sh` now adds the header only for `https://api.github.com/`. The loopback test asserts the token is withheld from a non-GitHub host; a recording fake `curl` asserts it reaches the API.
- P2, fixed: no CI job ran the installer harness. `app-check` (macOS runner) now runs `bash scripts/test-install-macos-app.sh`, and both installer files route to that job. The PR body said 6 cases; it is 5, corrected.
- P3, fixed: `install.sh` and `run.js` accept the redirect only from `https://github.com/7xuanlu/wenlan/releases/tag/` on a 3xx, strip query and fragment, reject malformed percent-escapes (the `URIError` no longer escapes the promise), and check every tag, pinned or resolved, against `vX.Y.Z[-pre]`. Live: `WENLAN_RELEASE_TAG='../../releases/download/v0.17.0'`, `v1.2.3?x=1` and `..` are refused with the tag format named; unpinned installs resolve `v0.17.0` on both paths.
- P3, fixed: `GH_TOKEN` now takes precedence over `GITHUB_TOKEN` (gh's order); HTTP 401 says the token is invalid or expired; the rate-limit copy says "wait for the limit to reset (up to an hour for anonymous callers)" instead of promising an hour.
- P3, fixed: the `self_update_check` cache comment names the release page, and the upgrade copy names `brew`, `npm` and `cargo`.
- P3, skipped: capturing response headers to tell primary from secondary rate limits. Both answers are the same for the user (wait, or use a token), so the extra parsing buys nothing at install time.
- P3, skipped: a shared redirect test matrix for `run.js`. The npm package has no test runner; the tag pattern was checked by hand for `v0.17.0`, `v0.18.0-rc.1` (accepted) and `v1.2.3?x=1`, `v1.2.3/`, `..`, `feature/test`, `0.17.0` (refused). Node 18 execution stays `unchecked` (Node 24 here); `node --check` is clean.

Receipts: harness 5 passed; mutation control (token to any `http*` URL) fails with `installer sent the token to a host that is not api.github.com: Bearer test-token`; `shellcheck -S warning`, `bash -n`, `node --check`, `actionlint` (pre-existing findings only) clean; `cargo test -p wenlan-mcp --lib self_update_check` 4 passed, fmt and clippy clean.

PR #608, opened 2026-08-26
fix(core): merge same-topic sibling shards before distill writes a page

### LINKED ISSUE #596: distill: same-topic sibling shards can become several unconfirmed pages after stepped re-splitting
## Problem

`cluster_distillation_rows` (PR #595) now re-splits an oversized same-space cluster in +0.05 threshold steps up to 0.92 instead of dropping it. That fixes the first-hour "no pages without a model" failure, but repeated tightening can slice one coherent topic into several near-duplicate shards (each ≥ 3 capture-seeded members), and nothing deterministic stops two shards of one topic from becoming two pages in the same pass:

- The Jaccard/subset overlap filter in `handle_distill` (`crates/wenlan-server/src/routes.rs`) and `synthesis/distill.rs` compares each cluster against *existing* pages, not against sibling clusters; disjoint shards have source Jaccard 0.
- `cap_document_majority_clusters` removes repeated document chunks, not semantic duplicates.
- New page births are `unconfirmed` (`post_write/page_create.rs`), and embedding/entity page matching only considers confirmed pages, so a page created from shard one cannot deduplicate shard two in the same pass.
- The LLM refinery (`refine_clusters_with_llm`) can merge shards, but the prompt's `split` action is silently treated as keep — only `merge` and `rename` are implemented (`prompts/defaults.rs` vs `synthesis/distill.rs`).
- Basic Memory mode returns pending clusters without LLM refinement; the distill skill checks each cluster's coherence individually, never siblings together.

Found by the adversarial review of #595 (P2). Acceptable for launch: the change replaces total data loss with bounded, reviewable, unconfirmed pages.

## Proposed fix

One of:

1. A cross-cluster review before writes: group pending siblings by space/entity and merge same-topic source sets (centroid cosine above the ceiling, or a shared majority of entities).
2. An embedding-only same-pass dedup path that also considers *unconfirmed* distilled pages created earlier in the pass.

Either way, implement the LLM `split` action or drop it from the prompt instead of silently keeping the cluster.

## Also from the same review (P3, not blocking)

- The empty-pass `hint` is generic; typed drop reasons from core (under floor, ceiling drop, covered by a page, unlinked) with route tests for each would let the skill say the real cause.
- The tuning loader does not range-check `formation_threshold` (`tuning.rs`); the loop now fails closed on a value that cannot advance, but a load-time `is_finite()` + `0.0..=1.0` check would reject the bad config outright.

7xuanlu:
## What

Closes #596. Stepped re-splitting (PR #595) can slice one topic into several near-duplicate shards, and nothing compared sibling clusters with each other: two shards of one topic became two unconfirmed pages in one distill pass. Pending clusters of one space are now merged when their centroids meet the re-split ceiling, before any page is written. The LLM refinery prompt no longer offers the `split` action, which the parser always treated as keep.

## How

- `merge_sibling_shards` (`crates/wenlan-core/src/db.rs`) runs inside `cluster_distillation_rows` after a bucket's base groups and every re-split descendant are emitted, before the size sort and `max_clusters` truncation. Both the server route and Basic Memory mode reach it through `find_distillation_clusters_scoped`.
- Two clusters of one bucket (one space, or the no-declared-space bucket) merge when their centroid cosine is at or above `RESPLIT_MAX_THRESHOLD` (0.92) and the merged estimated tokens stay within the token limit. The scan is in emission order; the earlier cluster absorbs the later one and is rebuilt through `build_distillation_cluster`, so its centroid, tokens and entity fields match any other cluster. The cross-space `Global` slice is untouched.
- The grouped cluster caps are formation-time bounds on how coarse the greedy grouping and the re-split ladder may be, not a ceiling on a real topic's size, so the merge does not check member count: a coherent 20-memory topic under cap 12 that the ladder cut into `[10, 10]` comes back as one cluster of 20. The token limit is the page-size guard. (The first cut kept the cap as a merge guard; the review showed that left the common case unfixed.)
- `crates/wenlan-core/src/prompts/defaults.rs` drops `split` from the refine-clusters prompt; `synthesis/distill.rs` says on the catch-all arm why. The entity signal is not used: a cluster carries one `entity_id` from its first member, not the member set a shared-majority rule needs.
- Three tests in `post_write/post_write_tests.rs` that set the process-global `WENLAN_MERGE_SHRINK_GUARD` now take the file's `env_lock()` like their neighbours; they flaked in parallel runs (`docs/ci-flake-policy.md`).
- Skipped on purpose: the issue's two P3 items (typed drop reasons, `formation_threshold` range check).

## Verification

- New tests: `same_topic_sibling_shards_merge_before_distill_writes_pages`, `ladder_shards_of_one_topic_rejoin_into_one_distill_cluster` (formation 0.60, cap 12, 20-memory topic, ladder emits `[10, 10]`, one cluster returned), `sibling_merge_refuses_a_union_over_the_token_limit`, `sibling_shards_in_different_spaces_are_not_merged`, `unrelated_distill_clusters_are_not_merged_as_siblings`, `refine_clusters_does_not_offer_the_unimplemented_split_action`. The three existing cap-asserting tests use unrelated subjects (centroids well below 0.92) and assert at least two clusters before the cap loop.
- `cargo test -p wenlan-core --lib distill` 71 passed; `sibling` 5; `cluster` 36; `threshold` 23; `formation` 2; `refine_clusters` 2; `prompt` 35. `cargo check -p wenlan-server`, `cargo fmt --check`, `cargo clippy -p wenlan-core --all-targets -- -D warnings`, `git diff --check` clean.
- Mutation controls: with the merge call disabled, `same_topic_sibling_shards_merge_before_distill_writes_pages` fails with `left: 2 right: 1`; with the member-cap check restored as `if union.len() > 12 { continue; }`, `ladder_shards_of_one_topic_rejoin_into_one_distill_cluster` fails with `… must come back as one cluster … [10, 10] left: 2 right: 1`. Raising each cap in the three cap tests collapses their fixtures to one blob (`[20]`, `[60]`, `[8]`), so the cap assertions are load-bearing.
- The full wenlan-core suite is `unchecked` locally; CI is the gate. No live daemon or model load beyond the cached BGE model the cap tests already used.

### DIFF
--- a/crates/wenlan-core/src/db.rs
+++ b/crates/wenlan-core/src/db.rs
@@ -2659,10 +2659,19 @@ const RESPLIT_MAX_THRESHOLD: f64 = 0.92;
 ///
 /// Two clusters of one bucket merge when their centroids' cosine is at or above
 /// `RESPLIT_MAX_THRESHOLD` — the same ceiling the ladder tightens toward, so a
-/// merge can only undo a split the ladder itself made. A merge that would push
-/// the result back over the bucket's member cap or over the LLM token limit is
-/// skipped: those two bounds are exactly why the ladder split in the first
-/// place, and re-merging past them would reintroduce the oversized cluster.
+/// merge can only undo a split the ladder itself made.
+///
+/// The member caps (`max_grouped_cluster_size` / `max_unlinked_cluster_size`)
+/// deliberately do NOT gate this merge. They are formation-time bounds: they
+/// tell the greedy grouping and the re-split ladder when a blob is still too
+/// coarse to be one topic, which is a statement about the threshold the blob
+/// was cut at, not a ceiling on how large a genuine topic may be. Once two
+/// shards' centroids meet the ladder's own ceiling, the evidence that they are
+/// one topic is stronger than the size heuristic that separated them, and a cap
+/// check here would just re-impose the split it exists to repair: a 20-memory
+/// topic under a cap of 12 emits [10, 10], and both halves would still become
+/// their own page. The LLM token limit is the real page-size guard, and it is
+/// still enforced below.
 ///
 /// Deterministic and model-free: candidates are scanned in emission order, the
 /// earlier cluster absorbs the later one, and the absorbing cluster is rebuilt
@@ -2672,7 +2681,6 @@ const RESPLIT_MAX_THRESHOLD: f64 = 0.92;
 fn merge_sibling_shards(
     memories: &[ClusterMemRow],
     clusters: Vec<DistillationCluster>,
-    cap: usize,
     token_limit: usize,
 ) -> Vec<DistillationCluster> {
     if clusters.len() < 2 {
@@ -2725,9 +2733,6 @@ fn merge_sibling_shards(
             // just a concatenation.
             let mut union = members[i].clone();
             union.extend_from_slice(&members[j]);
-            if union.len() > cap {
-                continue;
-            }
             let merged = build_distillation_cluster(memories, &union);
             if merged.estimated_tokens > token_limit {
                 continue;
@@ -2841,7 +2846,7 @@ fn cluster_distillation_rows(
         // about. Cross-space candidates (`Global`) are a different feature and
         // are left untouched.
         if matches!(mode, DistillationClusterMode::SpaceScoped) {
-            bucket = merge_sibling_shards(memories, bucket, cap, token_limit);
+            bucket = merge_sibling_shards(memories, bucket, token_limit);
         }
         clusters.extend(bucket);
     };
--- a/crates/wenlan-core/src/db.rs
+++ b/crates/wenlan-core/src/db.rs
@@ -2642,6 +2642,114 @@ const RESPLIT_STEP: f64 = 0.05;
 /// rather than tightened further.
 const RESPLIT_MAX_THRESHOLD: f64 = 0.92;
 
+/// Same-topic sibling shards (issue #596). The re-split ladder above can slice
+/// one coherent topic into several shards, and nothing downstream ever compares
+/// a pending cluster with its siblings: the Jaccard/subset filter in
+/// `handle_distill` scores a cluster only against *existing* pages, and two
+/// disjoint shards of one topic score 0 there. The page each shard creates is
+/// born `unconfirmed`, so shard one cannot deduplicate shard two inside the
+/// same pass either — one topic becomes two pages.
+///
+/// Merge those siblings back together before the pass writes anything, using
+/// the one same-pass signal already computed for every cluster: the mean-pooled
+/// `centroid_embedding`. (Entity identity is not usable here — a cluster
+/// carries a single `entity_id` taken from its first member, not the member set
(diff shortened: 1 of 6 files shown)


7xuanlu:

Verdict on `7975ad85`: MERGE, with one P2 and two P3s. All acted on in `25d1c970`.

**P2 — the member cap made the fix cosmetic.** The reviewer traced the ladder: a coherent 20-memory topic under cap 12 tightens to the 0.92 rung and emits `[10, 10]` with centroids about 0.94 apart, but the union (20) is over the cap, so the merge refused and the topic still became two pages; same for 30 → `[10, 10, 10]`. The guard only repaired the residue-loss geometry the first test used. Fixed: `merge_sibling_shards` no longer checks member count (the `cap` parameter is gone; the token guard stays), and the doc comment now describes the grouped caps as formation-time bounds, not a ceiling on a real topic's size. New regression `ladder_shards_of_one_topic_rejoin_into_one_distill_cluster` (formation 0.60, cap 12, one 20-memory topic; the ladder emits `[10, 10]`, the pass must return one cluster of 20). Mutation control with the cap check restored as `if union.len() > 12 { continue; }`: `a coherent topic the ladder cut in half must come back as one cluster … [10, 10] left: 2 right: 1`; reverted.

The three cap-asserting tests were reworked onto unrelated subjects so their emitted centroids stay well below 0.92, and each now asserts `clusters.len() >= 2` before the cap loop (two of them never asserted any cluster was returned). Probed by raising each cap: the fixtures collapse to one blob (`[20]`, `[60]`, `[8]`), so the cap is load-bearing; probes reverted.

**P3 — safety coverage.** Added `sibling_merge_refuses_a_union_over_the_token_limit` (halves at 1250 estimated tokens against a 1500 limit, union 2400, stays two; the fixture is asserted to qualify on similarity first) and `sibling_shards_in_different_spaces_are_not_merged` (near-identical centroids in spaces alpha and beta, one cluster per space).

**P3 — stale comments.** `eval/shared.rs` and four sites in `synthesis/distill.rs` no longer mention `split`.

**Beyond the review, per `docs/ci-flake-policy.md`:** the `threshold` filter went red twice with a different test each time (`merge_shrink_threshold_unset_returns_none`, `merge_shrink_threshold_valid_float`). Three tests in `post_write_tests.rs` set and remove the process-global `WENLAN_MERGE_SHRINK_GUARD` without the `env_lock()` their four neighbours already take. They now take it. Pre-existing; the new tests only perturbed scheduling.

Receipts after the fixes: `cargo fmt --check` and `cargo clippy -p wenlan-core --all-targets -- -D warnings` clean; `distill` 71, `sibling` 5, `cluster` 36, `threshold` 23, `formation` 2, `refine_clusters` 2 passed; `cargo check -p wenlan-server` clean; `git diff --check` clean. Full core suite `unchecked` locally; CI is the gate.

Reviewer points not acted on: the remaining greedy, emission-order sensitivity (a chain A~B, B~C, A≁C merges C only if the rebuilt A+B centroid reaches 0.92) and the first-member `entity_id` on a merged cluster are documented behaviour, not defects; `Global` mode returns one review-card candidate and is unchanged.

PR #610, opened 2026-08-26
fix(server): a blocked agent storing a memory gets 403, not 500

### LINKED ISSUE #586: fix(server): blocked agent storing a memory gets HTTP 500 instead of 403

7xuanlu:
## What

Closes #586. `POST /api/memory` from an agent whose store is blocked returned 500 (`Internal`) because the store handler wrapped `check_agent_for_write`'s `WenlanError::AgentDisabled` in `ServerError::Internal(e.to_string())`. The daemon already has `impl From<WenlanError> for ServerError` (`AgentDisabled` to 403, `Conflict` to 409, `NotFound` to 404, `Validation` to 422); that site just never used it.

## How

Every `.map_err(|e| ServerError::Internal(e.to_string()))` in `crates/wenlan-server/src/` (91 at the base) was traced to its `wenlan-core` callee, directly and through helpers it calls with `?`. Nine sites whose callee can raise one of the four mapped variants now use `.map_err(ServerError::from)`:

| Site | Callee | Code before the fix |
|---|---|---|
| `memory_routes.rs` store handler | `check_agent_for_write` | 500 instead of 403 (the issue) |
| `indexed_files_routes.rs`, `memory_routes.rs` | `db.update_memory` (via `apply_memory_update`) | 500 instead of 404 (memory gone) or 409 (concurrent edit) |
| `spaces_routes.rs` | `db.update_memory_space_opt` | same 404/409 masking |
| `spaces_routes.rs` | `create_space` | 500 instead of 422 (reserved name) |
| `spaces_routes.rs` | `update_space` | 500 instead of 404 or 422 |
| `spaces_routes.rs` | `delete_space` | 500 instead of 404 |
| `routes.rs` (two sites) | `db.clear_user_edited` (via `reject_page_draft_on_conn`) | 500 instead of 422 (page is a draft) |

The memory-correction handler (`handle_correct_memory`) also turned a missing target into a hand-written `ServerError::Internal("memory … not found")`. It now looks the memory up before the LLM-availability check and returns 404 (`{"error":"memory <id> not found"}`), with the DB error mapped through `ServerError::from`.

The other sites wrap callees that only raise `VectorDb`, `Io`, `Json` or `Embedding`, which map to 500 either way, so they are unchanged. `run_periodic_steep_with_api` is also unchanged: refinery phase errors are captured in `SteepResult.phases[].error` and the function returns `Ok`, so no `Validation` can reach its `map_err`.

Client effect: the CLI, the MCP server and the desktop app retry only connection failures, never HTTP statuses, so the new 4xx codes surface as the same immediate errors the 500s did, with the correct class. No status contract is documented anywhere in the repo.

## Verification

- `cargo fmt --check -p wenlan-server` and `cargo clippy -p wenlan-server --all-targets -- -D warnings` clean; `git diff --check` clean.
- New tests, all full-router (`build_router` + `tower::ServiceExt`): `memory_routes::blocked_agent_store_tests::store_from_disabled_agent_is_forbidden_not_internal_error` (asserts the body is exactly `Agent 'blocked-agent' is disabled`), `memory_routes::missing_memory_correction_tests::correct_missing_memory_is_not_found_not_internal_error`, `spaces_routes::not_found_mapping_tests::delete_missing_space_is_not_found_not_internal_error`.
- `cargo test -p wenlan-server --lib memory_routes -- --test-threads=1`: 41 passed; `spaces_routes`: 2 passed.
- Mutation controls: with the old `Internal` closure restored at the store handler, the 403 test fails with `left: 500, right: 403`; with the old LLM-first / `Internal`-on-`None` ordering restored in `handle_correct_memory`, the 404 test fails with `left: 500, right: 404`. Both restored and green.
- The ONNX `LoggingManager` panic seen once under the default parallel thread count did not reproduce: `cargo test -p wenlan-server --lib memory_routes` at default threads ran 3 times on this head (41 passed each) and 3 times on a detached checkout of main `81c4f073` (39 passed each), zero failures on either side, so the fixtures were left alone. The full server suite is `unchecked` locally; CI is the gate.

### DIFF
--- a/crates/wenlan-server/src/indexed_files_routes.rs
+++ b/crates/wenlan-server/src/indexed_files_routes.rs
@@ -75,7 +75,7 @@ pub async fn handle_update_chunk(
     };
     db.update_memory(&id, &req.content)
         .await
-        .map_err(|e| ServerError::Internal(e.to_string()))?;
+        .map_err(ServerError::from)?;
     Ok(Json(wenlan_types::responses::SuccessResponse { ok: true }))
 }
 
--- a/crates/wenlan-server/src/memory_routes.rs
+++ b/crates/wenlan-server/src/memory_routes.rs
@@ -383,7 +383,7 @@ pub async fn handle_store_memory(
         wait_at_store_lock_test_hook(StoreLockTestStage::AgentGate).await;
         db.check_agent_for_write(&resolved_agent)
             .await
-            .map_err(|e| ServerError::Internal(e.to_string()))?
+            .map_err(ServerError::from)?
     };
     let ambient_route_mode = wenlan_core::refinery::resolve_everyday(
         everyday_pin,
@@ -1835,7 +1835,7 @@ pub async fn handle_correct_memory(
     // Update the memory with corrected content
     db.update_memory(&id, &corrected)
         .await
-        .map_err(|e| ServerError::Internal(e.to_string()))?;
+        .map_err(ServerError::from)?;
 
     Ok(Json(serde_json::json!({
         "corrected": corrected,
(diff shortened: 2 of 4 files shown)


7xuanlu:

Verdict on `ae1dafbc`: MERGE, with two P2s and two P3s. All acted on in `ff18a68c`.

**P2 — an ordinary missing correction target still returned 500.** `handle_correct_memory` hand-converted `get_memory_detail` returning `None` into `Internal("memory … not found")`; the 404 added at the later `update_memory` site only covered a memory vanishing after the LLM call. Fixed: the lookup now runs before the LLM-availability check, `None` maps to `NotFound`, the DB error goes through `ServerError::from`. New full-router test `correct_missing_memory_is_not_found_not_internal_error` (`POST /api/memory/does-not-exist/correct` with no LLM configured returns 404 and `{"error":"memory does-not-exist not found"}`). Mutation control with the old ordering restored: `left: 500, right: 404`.

**P2 — the "unrelated ONNX flake" claim was unproven.** Measured: `cargo test -p wenlan-server --lib memory_routes` at the default parallel thread count, 3 runs on this head (41 passed each) and 3 runs on a detached checkout of main `81c4f073` (39 passed each), zero failures on either side. Per the reviewer's instruction the fixtures stay as they are.

**P3 — the 403 test only checked `contains("disabled")`.** It now asserts the body is exactly `Agent 'blocked-agent' is disabled`.

**P3 — the PR body's reason for leaving `run_periodic_steep_with_api` on 500 was wrong.** Corrected in the body: refinery phase errors are captured in `SteepResult.phases[].error` and the function returns `Ok`, so no `Validation` can reach that `map_err`.

Confirmed by the reviewer, no change needed: the literal sweep is exact (91 sites at the base, 82 after; every unchanged site's callee raises only `VectorDb`/`Io`/`Json`/`Embedding`); no client retries on HTTP status, so the new 4xx codes change classification only; the 403 body echoes the caller-supplied agent name and no stored reason; the variant prefix (`Validation error:`, `Agent disabled:`, …) drops from the client body text, which no handler or client parsed.

Receipts after the fixes: fmt, clippy (`-D warnings`), `git diff --check` clean; `memory_routes` 41 passed at `--test-threads=1`.

Issue #389, opened 2026-07-25
Daemon durability: fresh-boot total write loss (1x repro), unprotected BEGIN sites, and an acknowledged memory store vanishing within minutes


7xuanlu:
Three related durability findings from live-verify runs during the M3 wire-freeze work (branch `kg-m3-wire-freeze`, base main `7f3bd7ec`). All are pre-existing daemon behavior — the branch does not touch write paths; finding 1 was first seen on the PR #387 branch before these commits existed.

## Finding 1 — intermittent fresh-boot total write loss (2 occurrences)

Isolated daemon (port 7879, fresh `WENLAN_DATA_DIR`), writes via API (entity create, memory store, dual-write stub page). All writes API-visible (reads through the daemon return them). After SIGTERM: schema at user_version 94 persisted (incl. `pre_migration_*` backups), but `entities`, `pages`, `entity_page_map`, `memories` all EMPTY by row enumeration. No `-wal` file on disk — in the first occurrence (PR #387 verify, run 1 of its 4-run table) no `-wal` existed even while alive. Daemon log clean, no errors.

Fingerprint fits a **leaked open transaction on the single shared connection** (`MemoryDB` holds one `tokio::sync::Mutex<libsql::Connection>`): uncommitted writes are visible to the daemon's own connection (all API reads), invisible to external readers, produce no WAL append, and roll back on process death. Migrations commit at boot, so schema survives.

Not reproduced on demand: a 35-minute soak on an identical fresh boot (probes every 5 min: row enumeration + `BEGIN IMMEDIATE` write-lock probe from outside) stayed healthy — lock FREE throughout, entity + stub page survived SIGTERM. So the leak trigger is intermittent, and in the lossy runs it fired before the first API write (both lossy runs' first store landed ~60s after boot).

## Finding 2 — 14 production `BEGIN` sites with no ROLLBACK on error paths

Static sweep of `crates/wenlan-core/src/db.rs` (107 BEGIN sites; `db/page_map.rs` is clean). 14 production functions do `BEGIN -> fallible ops with ? -> COMMIT` where an error propagates without ROLLBACK, leaving the shared connection stuck in-transaction. Ranked by reachability:

Tier 1 (error swallowed by caller, scheduler/request-reachable — a leak here poisons the connection silently while the daemon keeps serving):
[...]

PR #611, opened 2026-08-26
fix(db): roll back when COMMIT fails so the writer connection is not left in a transaction

### LINKED ISSUE #593: db: a failed COMMIT in resolve_orphan_page_links leaves the writer transaction open
Found during the PR #590 review loop; out of scope there, filed separately.

## What

In `MemoryDB::resolve_orphan_page_links` (`crates/wenlan-core/src/db.rs`, the match on `result` just after the sweep body) a failed `COMMIT` returns through `?` without issuing `ROLLBACK`:

```rust
Ok(resolved) => {
    conn.execute("COMMIT", ()).await.map_err(|e| {
        WenlanError::VectorDb(format!("resolve_orphan_links commit: {e}"))
    })?;
    Ok(resolved)
}
Err(error) => {
    let _ = conn.execute("ROLLBACK", ()).await;
    Err(error)
}
```

If `COMMIT` fails with a retryable error (for example `SQLITE_BUSY` from a checkpoint or a reader holding the WAL), SQLite leaves the transaction open on the connection. The next `BEGIN` on the shared writer connection then fails with `cannot start a transaction within a transaction`, and every later write on that connection fails the same way until something issues `ROLLBACK`.

## Why it matters

The daemon has a single writer connection, so one failed `COMMIT` here can wedge all writes, not only the orphan sweep. The same shape (`COMMIT` via `?` with the rollback only on the body's `Err` arm) likely appears at other `BEGIN`/`COMMIT` sites in `db.rs`; #389 covered unprotected `BEGIN` sites but not a failed `COMMIT`.

## Suggested fix

On a `COMMIT` error, attempt `ROLLBACK` (ignoring its result) before returning the error, and sweep the other `COMMIT` sites for the same pattern. A regression can inject a `COMMIT` failure through a second connection holding a write lock.

7xuanlu:
## What

Closes #593. When `COMMIT` failed in `resolve_orphan_page_links`, the error was returned with the transaction still open on the single writer connection, so the next `BEGIN` failed with `cannot start a transaction within a transaction`. The same shape existed at 130 `COMMIT` sites across `wenlan-core`; four of them logged the failure and kept going inside the open transaction.

## How

- One helper, `commit_or_rollback(conn) -> libsql::Result<u64>` in `crates/wenlan-core/src/db.rs`: runs `COMMIT`, and on failure issues `ROLLBACK` before returning the original commit error. When the rollback itself fails and `!conn.is_autocommit()` (recovery failed, the connection is still inside a transaction) it logs at error level under the `[tx-watchdog]` tag; the harmless `no transaction is active` case stays quiet. It is private to `db.rs` (the `db/*` child modules reach it through `super::`) because the R4 drift guard forbids crate-visible `db.rs` functions that take a raw `libsql::Connection`. It returns the raw `libsql` error on purpose so every call site keeps its existing `.map_err(...)` and every log string stays byte-identical; a label parameter would have forced rewriting messages that are not of the `"{label}: {e}"` form.
- The 130 sites whose commit-failure path left the transaction open now call the helper: `db.rs` (112), `db/page_map.rs` (8), `db/migrations_v004_v009.rs` (6), and one each in `db/derived_artifact_sweep.rs`, `db/presence_review.rs`, `db/repair_page_rename.rs`, `db/repair_verification.rs`. `wenlan-server` has no `BEGIN`/`COMMIT`/`ROLLBACK` sites of its own.
- The four embedding loops that logged a failed commit and kept going (`m24 re-embed batch commit`, `m24 entity re-embed commit`, `null embed recovery commit`, `null entity embed recovery commit`) now propagate the error under that label and count progress only after the commit succeeds. Before, a persistent commit failure spun those loops forever while reporting progress. One consequence, kept on purpose: if the loop's `BEGIN` fails (still log-and-continue, issue #389), the following `COMMIT` errors with `no transaction is active` and the migration now aborts instead of carrying on with non-durable updates.
- Left alone: 24 production sites whose commit error arm already rolls back (including `rollback_repair_transaction` callers), and 10 test-module sites that `.unwrap()` fixture commits or pair a `commit()` with an explicit `rollback()` twin.
- Three source-text assertions in `db/repair_verification_test.rs` that pinned the literal `COMMIT` inside the verifier's finalization body now pin `commit_or_rollback(`; the order they encode (roll back the body error, then commit) is unchanged.

## Verification

- Fault injection: a busy second connection cannot make `COMMIT` itself fail in WAL mode (the first write fails instead; `db/presence_review_test.rs` documents that window). So the tests arm a `#[cfg(test)]` site-keyed `tokio::task_local!` hook (`with_failing_commit_at(site, fut)` and `arm_commit_failure(&conn, site)`, same mechanism as `repair_verification.rs`) that stages a deferred foreign-key violation inside the transaction; deferred constraints are checked at commit time, so `COMMIT` itself fails and the transaction stays active. Confirmed against SQLite directly: `FOREIGN KEY constraint failed (19)` on commit, then `cannot start a transaction within a transaction` on the next `BEGIN`.
- New test `a_failed_orphan_sweep_commit_does_not_wedge_the_writer_connection` (`db/main_tests.rs`): the commit error surfaces with its `resolve_orphan_links commit` label, and a following `insert_page` on the same connection succeeds.
- Mutation control, run twice: against the pre-fix code with only the hook and test added, and again with only the sweep's `COMMIT` reverted to a bare `conn.execute("COMMIT", ())` on the finished tree. Both fail with `a failed sweep commit must not wedge later writes: VectorDb("insert_page begin: SQLite failure: \`cannot start a transaction within a transaction\`")`; restored, green.
- New test `a_failed_recovery_commit_aborts_the_migration_instead_of_looping`: a `memories` row with no embedding, `run_migrations` again with the fault armed at `null_embed_recovery`; the error carries the loop's label and the connection is still usable afterwards. The fault is one-shot, so this proves the swallow (the mechanism), not the hang. Mutation control with that one site reverted to log-and-continue: `the staged commit failure must abort the migration: ()`; restored, green.
- `cargo fmt --check -p wenlan-core`, `cargo clippy -p wenlan-core --all-targets -- -D warnings`, `cargo check -p wenlan-server`, `git diff --check` clean; `scripts/m5-reader-sweep.py --update-inventory` unchanged.
- `cargo test -p wenlan-core --lib` filters after the follow-up: `orphan` 69, `migration` 179 (1 ignored), `commit` 46, `embed` 37, `r4_24` 5; before it also `repair` 238, `page_map` 39, `presence_review` 21, `derived` 14 — all passed.
- `unchecked`: the full wenlan-core lib suite (started, 331 passed, stopped for time) and `crates/wenlan-core/tests/` (`m4_community_gates` asserts two RAII-transaction functions contain no raw `BEGIN`/`COMMIT`/`ROLLBACK`; they were not touched). CI is the gate.
- Noted, not changed: the `#[cfg(test)]` `consume_repair_verification_test_commit_failure()` branch in `repair_verification.rs` returns before `COMMIT` without rolling back; test-only scaffolding, outside this issue.

### DIFF
--- a/crates/wenlan-core/src/db.rs
+++ b/crates/wenlan-core/src/db.rs
@@ -893,7 +893,7 @@ impl MemoryDB {
 
         match result {
             Ok(folded) => {
-                conn.execute("COMMIT", ())
+                commit_or_rollback(&conn)
                     .await
                     .map_err(|e| WenlanError::VectorDb(format!("fold commit: {e}")))?;
                 Ok(folded)
@@ -1017,7 +1017,7 @@ impl MemoryDB {
 
         match result {
             Ok(n) => {
-                conn.execute("COMMIT", ())
+                commit_or_rollback(&conn)
                     .await
                     .map_err(|e| WenlanError::VectorDb(format!("fold_entity commit: {e}")))?;
                 Ok(n)
@@ -1110,7 +1110,7 @@ impl MemoryDB {
 
         match result {
             Ok(n) => {
-                conn.execute("COMMIT", ())
+                commit_or_rollback(&conn)
                     .await
                     .map_err(|e| WenlanError::VectorDb(format!("retype commit: {e}")))?;
                 Ok(n)
@@ -1223,6 +1223,34 @@ fn take_fail_before_commit(page_id: &str) -> bool {
 #[inline(always)]
 fn take_fail_before_commit(_page_id: &str) -> bool {
     false
+}
+
+/// `COMMIT`, rolling the transaction back when the commit itself fails.
+///
+/// SQLite leaves the transaction open when `COMMIT` returns an error —
+/// `SQLITE_BUSY` raised by a checkpoint or by a reader still holding the WAL
+/// is the one seen in the field. `MemoryDB` owns ONE writer connection, so a
+/// caller that returns the commit error without rolling back leaves that
+/// connection mid-transaction: the next `BEGIN` fails with "cannot start a
+/// transaction within a transaction", and so does every write after it until
+/// something issues `ROLLBACK`. One failed commit would wedge every write in
+/// the daemon.
+///
+/// Drop-in for a bare `COMMIT` execute — same result, same error — so
+/// every call site keeps its own `map_err` label. Sites whose own
+/// commit-error arm already rolls back call `execute` directly and are left
+/// as they are.
+pub(crate) async fn commit_or_rollback(conn: &libsql::Connection) -> libsql::Result<u64> {
+    match conn.execute("COMMIT", ()).await {
+        Ok(rows) => Ok(rows),
+        Err(error) => {
+            // Best effort. If the failure did end the transaction after all,
+            // this reports "cannot rollback - no transaction is active",
+            // which is the state we wanted anyway.
+            let _ = conn.execute("ROLLBACK", ()).await;
+            Err(error)
+        }
+    }
 }
 
 /// One recorded version of a page, as stored in `page_history`.
@@ -4767,7 +4795,7 @@ impl MemoryDB {
                 .await
                 .map_err(|e| WenlanError::VectorDb(format!("set user_version=2: {}", e)))?;
 
-            conn.execute("COMMIT", ())
+            commit_or_rollback(&conn)
                 .await
                 .map_err(|e| WenlanError::VectorDb(format!("migration2 commit: {}", e)))?;
 
@@ -4829,7 +4857,7 @@ impl MemoryDB {
                 .await
                 .map_err(|e| WenlanError::VectorDb(format!("set user_version=3: {}", e)))?;
 
-            conn.execute("COMMIT", ())
+            commit_or_rollback(&conn)
                 .await
                 .map_err(|e| WenlanError::VectorDb(format!("migration3 commit: {}", e)))?;
 
@@ -4962,7 +4990,7 @@ impl MemoryDB {
                 .await
                 .map_err(|e| WenlanError::VectorDb(format!("set user_version=10: {}", e)))?;
 
-            conn.execute("COMMIT", ())
+            commit_or_rollback(&conn)
                 .await
                 .map_err(|e| WenlanError::VectorDb(format!("migration10 commit: {}", e)))?;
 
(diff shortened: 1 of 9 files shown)


7xuanlu:

Verdict on `0058cfac`: MERGE, with one P2 and one P3. Both acted on in `69577ab5`.

**P2 — four embedding loops could spin forever and report false progress after a persistent commit failure.** The `m24` re-embed batch and entity loops and the two null-embedding recovery loops logged a failed commit and continued. Each now propagates `commit_or_rollback`'s error under its existing label (`m24 re-embed batch commit`, `m24 entity re-embed commit`, `null embed recovery commit`, `null entity embed recovery commit`) and counts progress only after the commit succeeded. New test `a_failed_recovery_commit_aborts_the_migration_instead_of_looping`, driven through a site-keyed generalisation of the task-local fault hook (`with_failing_commit_at` / `arm_commit_failure`; the orphan-sweep test moved onto it). Mutation control with one site reverted to log-and-continue: `the staged commit failure must abort the migration: ()`; restored, green.

Noted in the body as a conscious behaviour change: when a loop's `BEGIN` fails (still log-and-continue, issue #389), the following `COMMIT` errors with `no transaction is active` and the migration now aborts instead of carrying on with non-durable updates.

**P3 — a real rollback failure was indistinguishable from the harmless "no transaction active" case.** The commit error still returns unchanged; the rollback result is now captured and, when it failed and `!conn.is_autocommit()`, logged at error level under the `[tx-watchdog]` tag (the same tag `assert_autocommit_or_recover` uses).

Confirmed by the reviewer, no change needed: the sweep of 130 sites is exact; the 24 sites left alone already roll back; the test hook cannot fire outside `cfg(test)`; no P1.

Receipts after the fixes: fmt, clippy (`-D warnings`), `cargo check -p wenlan-server`, `git diff --check` clean; `scripts/m5-reader-sweep.py --update-inventory` unchanged; `cargo test -p wenlan-core --lib` filters `orphan` 69, `migration` 179 (1 ignored), `commit` 46, `embed` 37, `r4_24` 5, all passed. The full core suite stays `unchecked` locally; CI is the gate.

7xuanlu:
CI `linux-nextest` failed on `r4_test_support_raw_and_api_manifests_match_exactly`: the drift guard forbids a crate-visible `db.rs` function whose signature carries a raw `libsql::Connection`. `ec0a10cf` makes `commit_or_rollback` private; the `db/*` child modules still reach it through `super::`. Guard test, `cargo check -p wenlan-core -p wenlan-server` and fmt green locally.

PR #609, opened 2026-08-26
fix(mcp,core): trim MCP recall hits, render brief and lint as text, no migration backups on a fresh store

7xuanlu:
## What

Soak gaps 6 and 7 from the 2026-08-25 usefulness pass (`wenlan-usefulness-soak-2026-08-25.md`, gaps table).

- MCP `recall` (and the Brief's related context) returned the daemon's full `SearchResult` (about 30 fields per hit: chunk index, raw score, source text, content hash, stability, quality, language and other ingest internals). The agent now sees a `RecallHit` with the fields the skills read.
- MCP `brief` returned raw JSON; it now renders text.
- MCP `lint` returned the typed report as JSON text, which no skill reads. General now returns the text `wenlan lint` prints; Deep returns the typed report as `structured_content`. The two are not combined: Claude Code and Codex forward `structuredContent` and drop text blocks when both are set (code.claude.com/docs/en/agent-sdk/custom-tools: "Text blocks in content are not forwarded"), so a text-plus-structured result would never have been model-visible.
- A fresh install replays every migration onto an empty store and took a pre-migration backup at each backed-up migration: 29 `pre_migration_<n>_backup.db` files, 37 MB, beside a 5 MB live database, with nothing to restore.

## How

- `crates/wenlan-mcp/src/tools.rs`: `RecallHit` (source_id, title, content, score, last_modified, memory_type, space, source_agent, confidence, confirmed, entity_name, supersedes, version, pending_revision, merged_from, last_delta_summary), built from `&SearchResult`; `recall` and the Brief's related context serialize it. `brief` renders `Space Brief: {space} (v{version})`, the last-session line, then Active and Backlog items as `- {id} (v{version}): {text} [gate: {gate}]`, so `/handoff` still has the ids and versions it reconciles. `lint` returns `CallToolResult::success([text(report.render_text())])` for General and `CallToolResult::structured(report)` for Deep; both lint `SKILL.md` files describe the split.
- `crates/wenlan-types/src/lint_report.rs`: `LintReport::render_text` plus the helpers, moved verbatim from `crates/wenlan-cli/src/commands/lint.rs`, which now prints `report.render_text()`.
[...]

PR #612, opened 2026-08-26
fix(lint): a fresh store without a model source is not a failing lint

7xuanlu:
## What

Gauntlet finding 15 and soak gap 4. On a fresh store with no model source chosen, `wenlan lint` reported 56 checks / 54 passed, two findings (`kg.substrate_liveness`, `serving.channel.graph`) and exit 1, with only machine codes to explain them. There was nothing to check yet: background enrichment is paused until the owner picks a model source, and the capture path already says so.

## How

- Readiness signal: `model_source_configured` = `EverydaySource::parse(load_config().everyday_source)` is `Some`. That is the same input `refinery::resolve_everyday` reads to produce the capture hint "background enrichment is paused until you choose a model source", so lint and capture agree. The existing `provider_ready` latch was not used: it is a process-wide "some provider has served traffic" flag that is false on a freshly started daemon even with a source configured. A source that is chosen but currently unavailable keeps the pin true, so nobody is told to choose a source they already chose.
- The flag is threaded into `KgRunConfig` and `ServingRunConfig` only. It joins the config fingerprint as `LintConfigSetting::ModelSourceConfigured` (byte id 27) so reports before and after the setting differ.
- With no source configured, `kg.substrate_liveness` and `serving.channel.graph` are `outcome: pass`, `applicability: expected_empty`, `precondition: expected_empty`, `summary_code: expected_empty`, no `recommendation_code`, and a new `action_code: choose_model_source`. That combination was already legal in the check legality table; `LintPrecondition::ExpectedEmpty` simply had no producer.
- `action_code` is a new optional field on `LintCheckResult` holding a one-variant `LintActionCode`, `#[serde(default, skip_serializing_if = "Option::is_none")]`. `LintRecommendationCode` keeps its four released values, so a CLI or MCP one patch behind still parses a report from this daemon (older readers ignore unknown keys). `LintCheckResultInput` is untouched; the field is read on `LintCheckResultWire` beside `gate_effect`, and producers opt in with `with_action_code`.
- `crates/wenlan-types/src/lint_catalog.rs` gains one plain sentence per summary, recommendation and action code; the CLI renders each check as code plus sentence, and prints these expected-empty checks under a `Waiting on you (N):` section instead of hiding them among passes. Both lint `SKILL.md` files carry the same code-to-sentence tables, and `plugin_distribution.rs` asserts the tables match the catalog so the two copies cannot drift.
- Exit code: 0 unless there are actionable findings (1) or the report is incomplete (2).
- A related bug fixed on the way: the kg liveness assessment reported `truncated=true` with zero evidence. `Assessment::reports_evidence` now says whether an assessment can point at records at all (`structural` true, `inventory_with_basis` and `liveness` false). No global `LintCoverage` rule was added: three producers (`identity/result.rs`, `operations/result.rs`, `semantic.rs`) legitimately emit `truncated=true` with no evidence, and `semantic_test.rs` asserts one of them.
- No schema or catalog version bump: no check ids changed, the wire change is additive, nothing persists a `LintCheckResult`, and bumping `LINT_REPORT_SCHEMA_VERSION` would re-point `PREVIOUS_LINT_REPORT_SCHEMA_VERSION` in `repair.rs` and invalidate persisted repair manifests.
- `CLI_READERS` in `truth_manifest.rs` gains the `wenlan sweep` row clap already had, so the `wenlan` bin-target manifest test passes again.

## Verification

- `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean for wenlan-types, wenlan-core and wenlan (the CLI package). `git diff --check` clean.
- `cargo test -p wenlan-types`: 185 passed (new: `a_frozen_schema_five_reader_still_parses_the_fresh_store_check`, `a_check_without_an_action_does_not_grow_a_wire_key`). `cargo test -p wenlan-core --lib lint`: 239 passed, 2 ignored. `cargo test -p wenlan`: all 11 targets green, including the bin-target `truth_manifest_cli_rows_match_clap_subcommands`. `cargo test -p wenlan --lib`: 57 passed (four new: `a_fresh_store_without_a_model_source_exits_zero`, `exit_code_is_one_only_for_actionable_findings_and_two_for_incomplete`, `the_model_source_checks_render_with_a_plain_sentence_and_no_findings`, `a_rendered_finding_carries_both_the_code_and_the_plain_sentence`). `cargo test -p wenlan --test lint_cli`: 19 passed; `--test cli_integration`: 40 passed.
- Mutation controls (break confirmed with `git diff`, then restored): putting `ChooseModelSource` back into `LintRecommendationCode` fails the frozen-reader test with `unknown variant \`choose_model_source\`, expected one of \`review_finding\`, \`restore_prerequisite\`, \`rerun_after_snapshot_stabilizes\`, \`inspect_runtime\``; reverting the liveness reclassification fails `no_model_source_is_a_complete_expected_empty_pass_that_says_to_choose_one` with `left: Finding, right: Pass`; dropping the `reports_evidence` guard fails `aggregate_only_liveness_never_claims_truncated_evidence` with `assertion failed: !liveness.coverage().truncated()`.
- Live, rerun after the wire change on `WENLAN_PORT=17932`: raw `GET /api/lint` emits no `recommendation_code` outside the released set and `action_code` on exactly 2 of 56 checks; `wenlan lint` prints `action: choose_model_source` under `Waiting on you (2)`. First run: an isolated daemon (`WENLAN_PORT=17931`, scratch `WENLAN_DATA_DIR`, `WENLAN_NO_AUTOSTART=1`, scratch `knowledge_path`, no `everyday_source`). Two captures returned `"enrichment": "paused"` with the choose-a-source hint; `wenlan lint` then exited 0 with `Lint: 56 checks, 56 passed, 0 actionable findings, 0 advisories, 0 incomplete` and both checks under `Waiting on you (2)` with the sentence "There was nothing here to check, which is the expected state right now. Run `wenlan setup` and choose a model source so Wenlan can build this in the background." The developer daemon on 7878 was untouched.
- `unchecked`: CI does not run the `wenlan` bin-target unit test, so the manifest check that was failing on main went unnoticed there; adding `--bin wenlan` to the canonical test inventory is left as a follow-up. Fake `HOME` was not set for the live run (the worktree guard refuses it); every home-touching startup path was checked to be a no-op on this machine instead.

### DIFF
--- a/crates/wenlan-cli/src/commands/lint.rs
+++ b/crates/wenlan-cli/src/commands/lint.rs
@@ -155,6 +155,19 @@ fn render_human(report: &LintReport) -> String {
     }
     append_findings(&mut output, "Findings", report, LintGateEffect::Actionable);
     append_findings(&mut output, "Advisories", report, LintGateEffect::Advisory);
+    // A check that passed but still carries advice is waiting on the owner, not
+    // broken — e.g. a graph channel with no model source chosen. It is not a
+    // finding and does not change the exit code, but the reader still has to see
+    // the one thing they can do about it.
+    output.push_str("Waiting on you");
+    let waiting: Vec<_> = report
+        .checks()
+        .iter()
+        .filter(|check| {
+            check.outcome() == LintOutcome::Pass && check.recommendation_code().is_some()
+        })
+        .collect();
+    append_selected(&mut output, &waiting);
     output.push_str("Incomplete");
     let incomplete: Vec<_> = report
         .checks()
@@ -198,6 +211,18 @@ fn append_selected(output: &mut String, checks: &[&wenlan_types::lint::LintCheck
             )),
             None => output.push_str(&format!("  {}: {summary}\n", check.check_id())),
         }
+        // The codes above are stable identifiers for scripts. This line is the
+        // same thing in plain English for the person reading the terminal.
+        output.push_str(&format!(
+            "    {}{}\n",
+            check.summary_code().meaning(),
+            check
+                .recommendation_code()
+                .map_or_else(String::new, |recommendation| format!(
+                    " {}",
+                    recommendation.action()
+                ))
+        ));
         let affected = check.metrics().iter().find_map(|metric| {
             if metric.code() == LintMetricCode::AffectedRecords {
                 match metric.value() {
@@ -302,6 +327,7 @@ const fn recommendation_name(recommendation: LintRecommendationCode) -> &'static
         LintRecommendationCode::RestorePrerequisite => "restore_prerequisite",
         LintRecommendationCode::RerunAfterSnapshotStabilizes => "rerun_after_snapshot_stabilizes",
         LintRecommendationCode::InspectRuntime => "inspect_runtime",
+        LintRecommendationCode::ChooseModelSource => "choose_model_source",
     }
 }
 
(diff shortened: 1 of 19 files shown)


7xuanlu:

Verdict on `1c12590f`: MERGE, with one P2 and two P3s. All acted on in `0a70aae6`.

**P2 — a new `recommendation_code` value under the same schema version broke patch-compatible old readers.** `ChooseModelSource` was a fifth serialized `LintRecommendationCode` while the report still declared schema 5, and a CLI or MCP one patch behind (explicitly compatible per `version_check.rs`) could not parse it. Fixed as prescribed: the enum is back to its four released values; the ask rides in a new optional `action_code` field (`#[serde(default, skip_serializing_if = "Option::is_none")]`, one-variant `LintActionCode`), read on `LintCheckResultWire` beside `gate_effect` so `LintCheckResultInput` keeps the shape an old reader parses. New tests: `a_frozen_schema_five_reader_still_parses_the_fresh_store_check` (a test-local copy of the base-`43c00470` reader with only the four variants parses the new check) and `a_check_without_an_action_does_not_grow_a_wire_key`. Mutation control with the variant restored: `unknown variant `choose_model_source`, expected one of `review_finding`, `restore_prerequisite`, `rerun_after_snapshot_stabilizes`, `inspect_runtime``; reverted, green. Live raw `GET /api/lint` on a scratch daemon: no `recommendation_code` outside the released set, `action_code` on 2 of 56 checks.

**P3 — MCP exposed codes without the promised plain explanation.** Both lint `SKILL.md` files now carry a "Say what each code means" section: tables for all six `summary_code`, four `recommendation_code` and the one `action_code` value, plus the rule that a passing check carrying an action is waiting on the owner and is not an actionable finding. A new `plugin_distribution.rs` test reads the sentences off the catalog and asserts both files contain each code with its exact sentence, so the two copies cannot drift. No `summary_text`/`recommendation_text` fields were added: PR #609 makes General MCP lint return the CLI text.

**P3 follow-up — `wenlan sweep` missing from `CLI_READERS`.** The manifest row is added; `cargo test -p wenlan --bin wenlan` now passes (`truth_manifest_cli_rows_match_clap_subcommands ... ok`), where it exited 101 before. Adding `--bin wenlan` to the canonical CI test inventory is a deliberate skip, left as a follow-up.

Receipts after the fixes: fmt, clippy (`-D warnings`) and `git diff --check` clean for wenlan-types, wenlan-core and the CLI; `cargo test -p wenlan-types` 185 passed; `cargo test -p wenlan-core --lib lint` 239 passed, 2 ignored; `cargo test -p wenlan` all 11 targets green; `cargo check --all-targets` for the five daemon crates clean.

7xuanlu:
CI `linux-nextest` and `macos-nextest` failed on `truth_manifest::tests::cli_table_equals_inventory_document`: the new `wenlan sweep` row in `CLI_READERS` also has to appear in `crates/wenlan-core/contracts/m5-reader-manifest-inventory.md`, which is compiled into that test. `d4845514` adds the row and updates the count to 23 in the document and the manifest's doc comment. `cargo test -p wenlan-core --lib truth_manifest` (16 passed) and the `wenlan` bin manifest test are green locally.

PR #600, opened 2026-08-26
fix(core,cli): stop the fresh-install daemon crash-loop under launchd

7xuanlu:
Pre-launch tracker row 10 (`wenlan-audit-2026-08-22.md`), first-run gauntlet finding F4 plus the rest of F8.

## What was wrong

Every fresh macOS install in the gauntlet crash-looped: `wenlan background on` printed `Installed and started com.wenlan.server.` while launchd restarted the daemon every 10 s with exit 1 and `stderr path = /dev/null`. The daemon's own log showed each attempt ending in `Failed to retrieve model_optimized.onnx`.

Mechanism (confirmed by the gauntlet's A/B replay): on a fresh account `resolve_fastembed_cache_dir` finds no populated cache and returned `None`, so fastembed used its own default — `.fastembed_cache` **relative to the process working directory** (`FASTEMBED_CACHE_DIR` overrides it; the old comment claiming `~/.fastembed_cache` was wrong). launchd starts the daemon at `/`, where the 210 MB first-run download cannot be written. Channels that start the daemon from a writable cwd (brew archive, systemd, schtasks) worked by accident and left a `.fastembed_cache` wherever the daemon happened to run.

## What changed

| Area | Change |
|---|---|
| `wenlan-core` `db.rs` | `MemoryDB::new` always passes an explicit cache directory: `daemon_fastembed_cache_dir(db_path)` returns a populated cache from the resolver, else `FASTEMBED_CACHE_DIR` (CI pre-populates it), else `<db_path>/fastembed_cache`, created before the embedder thread starts. The log line always names the directory now. |
| `wenlan` CLI `service.rs` | `background on` waits up to 10 s for `/api/health` on every platform. It prints "Installed and started …" only once the daemon answers. On macOS a daemon that stopped during this start fails the command with the error it logged after the command began ("installed …, but the daemon stopped with an error (<log>): …"; daemon log first, then launchd's stderr file; an error left over from an earlier run is never blamed on this start). Otherwise it says the daemon is still starting, that a first start downloads the 210 MB embedding model, and where to look. The launchd plist sends stderr to `<data root>/logs/launchd-stderr.log` instead of `/dev/null`; launchd does not rotate that file, and the daemon writes only bootstrap failures there (progress bars are hidden off a terminal). |
| `wenlan` CLI `setup.rs` | `wenlan doctor` lists the launchd stderr log; the log-reading helpers are shared with `background on`. |
| `crates/wenlan-cli/README.md` | Documents the wait and the error report. |

[...]

PR #613, opened 2026-08-26
ci(gauntlet): expect the health-verified background-on line

7xuanlu:
## What

The first-run gauntlet still asserted the pre-#600 `wenlan background on` line, `Installed and started com.wenlan.server.` (trailing period). Since PR #600 the CLI waits for the daemon to answer before it claims a start, and on success prints `Installed and started com.wenlan.server; daemon healthy at http://127.0.0.1:7878.`. The three scripts that check that line (`linux-install.sh`, `plugin-setup.sh`, `npx-setup.sh`) now expect the health-verified form, `Installed and started com.wenlan.server; daemon healthy at`, so the assertion also proves the health wait ran.

## Why

The v0.17.1 gauntlet (run 32981419005) came back red on `linux-install` (both Ubuntu runners), `plugin-setup` (both runners) and `npx-setup` with the daemon healthy in 1–3 s on every one of them — the only failing row was this stale substring. The macOS app channel's `plist-server-exists` failure is a real product finding and is left untouched.

## Verification

- `bash -n` on the three scripts.
- The v0.17.1 findings show the exact new line on all five red jobs, e.g. `got: Installed and started com.wenlan.server; daemon healthy at http://127.0.0.1:7878.`
- Will be proven by re-dispatching `first-run-gauntlet.yml` against `v0.17.1` after merge.

### DIFF
--- a/scripts/first-run/linux-install.sh
+++ b/scripts/first-run/linux-install.sh
@@ -69,7 +69,7 @@ fi
 export PATH="$BIN_DIR:$PATH"
 info path-export "export PATH=\"$BIN_DIR:\$PATH\""
 check_output setup-basic "Wenlan is set up for local memory." -- wenlan setup --basic
-check_output background-on "Installed and started com.wenlan.server." -- wenlan background on
+check_output background-on "Installed and started com.wenlan.server; daemon healthy at" -- wenlan background on
 check unit-file-exists -- test -f "$UNIT"
 check_output unit-enabled "enabled" -- systemctl --user is-enabled wenlan-server
 wait_health "$HEALTH" 240 || true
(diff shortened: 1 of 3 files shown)


7xuanlu:
## Adversarial review (Codex Sol, gpt-5.6-sol, one pass)

Verdict: **MERGE**, no P1 or P2 findings.

1. `{installed}` is the constant `com.wenlan.server` on macOS and Linux; Linux converts the label only for the unit filename, and the npx path invokes the same CLI (`crates/wenlan-cli/src/commands/service.rs`, `crates/wenlan-cli/npm/run.js:175-179`).
2. No checked job sets `WENLAN_BIND_ADDR`, so the CLI defaults to `127.0.0.1:7878` and never takes the plain-output branch that the old string matched.
3. The npx CLI runs with inherited stdio and the outer check captures combined output into `$LOG`, so `grep -F` still sees the line; `background-on-again` stays a bare rc check because health and version are verified right after it.
4. No exact old string remains anywhere under `scripts/first-run/`; `brew.sh:66` is a bare rc check and the Windows prefix checks already match the health-suffixed output.

PR #614, opened 2026-08-27
fix(app): register the daemon with launchd first; the sidecar is the fallback

7xuanlu:
## What

On a fresh Mac the app now registers the daemon with launchd first and starts its own sidecar only when that registration is skipped or fails. Before, `setup()` spawned the sidecar at once and ran the first-run `wenlan background on` in parallel, and the two fought for port 7878 (first-run gauntlet finding F16, tracker row 23):

- v0.17.1 run 32981419005: `[first-run] wenlan background on failed: … Error: request graceful daemon shutdown before reinstall` — the CLI's pre-install shutdown request hit the still-booting sidecar, so `com.wenlan.server.plist` was never written.
- v0.17.1 run 33003051324: `background on` won, but launchd's daemon found the sidecar on the port and looped on `Existing healthy daemon on port 7878 — exiting 75 (launchd retry)`; `launchctl print` showed `runs = 1`, `last exit code = 1`, `state = spawn scheduled`.

Both runs at 0.17.0 failed the same check; the crash-loop (F4) had hidden it.

## How

- `app/src/lib.rs`: the macOS sidecar spawn moves from `setup()` into the completion of the first-run install task. Once the install has settled, `decide_startup_sidecar(preflight_ok, current_server_plist_matches_selected_data_dir())` picks the owner: launchd when the server LaunchAgent targets the selected data root, otherwise the sidecar (opted out, unstable app path, isolated data dir, or a failed `background on`). Windows and Linux keep the immediate spawn.
- `app/src/daemon_start.rs`: `StartupSidecar` + `decide_startup_sidecar` (pure, tested); a `LAUNCHD_INSTALL_PENDING` flag so the on-demand "Start Wenlan" command does not spawn a rival sidecar while the job is being registered (`decide_daemon_start` gains the input; the UI sees it as `launchd_managed`).
- `app/src/lifecycle.rs`: `set_run_at_login(true)` stops an app-owned sidecar (`daemon_start::stop_sidecar`, a no-op when launchd owns the daemon) before `wenlan background on`, since the toggle hit the same refusal.
- `docs/cross-platform.md`: one paragraph on the ownership order.

## Verification

- `cargo test -p wenlan-app --lib`: full suite (see the CI note below for the local run).
- New tests: `pending_launchd_install_defers_the_sidecar`, `startup_sidecar_is_the_fallback_owner`; the four existing `decide_daemon_start` tests carry the new input.
- `cargo clippy -p wenlan-app --all-targets -- -D warnings` clean, `cargo fmt` clean.
- Real proof needs a published build: after 0.17.2 is cut, the gauntlet's `macos-app` channel must pass `plist-server-exists`, `plist-desktop-exists`, and `launchctl-server-loaded` with the daemon run by launchd, and `daemon-after-app-quit` must then report the daemon still reachable (service-owned by design).

### DIFF
--- a/app/src/daemon_start.rs
+++ b/app/src/daemon_start.rs
@@ -14,7 +14,7 @@
 //! and [`stop_sidecar`] ends it on quit, and on Windows the process is bound
 //! to a kill-on-close job object so a hard kill of the app takes it too.
 
-use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicU64, AtomicUsize, Ordering};
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 use tauri_plugin_shell::process::CommandChild;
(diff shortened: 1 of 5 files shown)


7xuanlu:
## Adversarial review (Codex, gpt-5.6-sol, one pass) and what changed

The reviewer's verdict on the first commit was DO NOT MERGE, with two P1s and two P2s. All four are acted on in the second commit (`fix(app): launchd owns the daemon only with a loaded job; lock the owner decision`).

**P1 — the owner test trusted the plist file alone.** `wenlan background on` writes `com.wenlan.server.plist` before `launchctl load` (service-manager 0.11 `launchd.rs` `install`), so a failed load leaves a matching file with no job; the app then skipped its sidecar and the "Start Wenlan" button, reading the same file, refused too. Fixed: `lifecycle::launchd_owns_server_daemon` requires the file *and* the label in `launchctl list` (the same `label_is_loaded` the Run at Login toggle already uses). Both the startup decision and the button use it. Regression `launchd_owns_the_daemon_only_when_its_job_is_loaded`; mutation control (predicate reverted to the file-only check) fails it with `a matching plist that launchd never loaded must not count as launchd's daemon`.

**P1 — the pending flag did not close the race.** Setup cleared the flag before deciding and spawning, and the button sampled the flag after its own awaited health probe, so both could spawn; the second spawn kills the first, and a second child that had seen the first one healthy exits 75 (zero owners). Fixed: `daemon_start::OWNER_DECISION` serialises every owner decision with the spawn it leads to (`settle_startup_owner` for startup, `start_daemon_under_owner_lock` for the button), the pending flag is a counted RAII guard (`LaunchdInstallPending`) released on drop — a panicked install can no longer pin the button on `launchd_managed`, and the Run at Login toggle holds it as well — and `decide_daemon_start` gains `sidecar_alive`: a live child of ours is reported as `Started`, never spawned twice. Tests `a_live_sidecar_is_never_spawned_twice`, `pending_flag_counts_overlapping_installs_and_releases_on_drop`.

**P2 — Run at Login handover failure leaves no daemon.** The `set_run_at_login` command now calls `start_daemon_if_unowned` when enabling fails, so the daemon comes back as a sidecar unless something owns it.

**P2 — `[init]` health loop budget.** Unchanged by this PR: the loop waits about 102 s in total (200 ms doubling, ten attempts), which covers the install (2 s shutdown request, install, 10 s first-health wait) plus a sidecar boot; a first boot that downloads the model exceeded it before as well. Not changed here.

Reviewer's note that tests/clippy were unchecked in its read-only sandbox: locally `cargo test -p wenlan-app --lib` is 447 passed, 0 failed, 1 ignored; `cargo clippy -p wenlan-app --all-targets -- -D warnings` and `cargo fmt` clean. Frontend needs nothing: `DaemonStartResult` keeps its four variants (`src/lib/tauri.ts`).

PR #620, opened 2026-08-28
fix(gauntlet): wait for the launch agent plists instead of testing once

### LINKED ISSUE #619: gauntlet: plist-desktop-exists races the app's first-run install (F17)
Seen on the v0.17.3 full gauntlet run https://github.com/7xuanlu/wenlan/actions/runs/33124448376: the `macos-app` channel failed only `plist-desktop-exists` (54 PASS, 2 FAIL reported, one check). The rerun with the same binaries, https://github.com/7xuanlu/wenlan/actions/runs/33124875152, passed 56/0.

Mechanism: `scripts/first-run/macos-app.sh:74` tests for `~/Library/LaunchAgents/com.wenlan.desktop.plist` immediately after `wait_health` returns. The app writes that file only after `wenlan background on` returns (`app/src/lifecycle.rs` around lines 761-770: `install_server_plist_via_subprocess`, then `install_app_plist`), and `background on` returns only once it has seen the daemon healthy — the same signal the gauntlet polls. When the gauntlet's poll wins the race (3 s to health on the failed run), the check runs a few hundred milliseconds early.

Fix: poll for the plist for up to 30 s, the way `app-process-started` polls for the process at line 71. Gauntlet script only; nothing in `app/` changed between v0.17.2 and v0.17.3 except the version bump.

7xuanlu:
Closes #619.

## The race

`scripts/first-run/macos-app.sh` tested for `~/Library/LaunchAgents/com.wenlan.desktop.plist` the instant `wait_health` returned. The app writes that file only after `wenlan background on` comes back (`app/src/lifecycle.rs`, `install_server_plist_via_subprocess` then `install_app_plist`), and `background on` itself returns once it has seen the daemon healthy — the same signal the gauntlet polls. When the gauntlet's poll wins, the check runs a few hundred milliseconds early.

Evidence: on v0.17.3, run [33124448376](https://github.com/7xuanlu/wenlan/actions/runs/33124448376) failed `plist-desktop-exists` at `seconds-to-health 3`, and rerun [33124875152](https://github.com/7xuanlu/wenlan/actions/runs/33124875152) passed 56/56 with the same binaries at `seconds-to-health 15`. Nothing under `app/` changed between v0.17.2 and v0.17.3 beyond the version bump.

## The change

`lib.sh` gains `check_eventually NAME SECS -- CMD ARGS...`: it retries the command once a second for up to `SECS` and records exactly one row — `ready after Ns` on PASS, the last attempt's output on FAIL. It runs each attempt through the existing `_gauntlet_run`, so the per-check log file and `GAUNTLET_TIMEOUT` behave as before.

Three checks in `macos-app.sh` use it: both plist tests, and `app-process-started`, which had hand-rolled the same loop inside a `bash -c` string. Generalizing the helper rather than special-casing the one flaky line is the point — the next check that races a background writer now has somewhere to go.

`first-run-gauntlet.yml` runs `lib.test.sh` in the `resolve` job before the channels fan out. These unit tests existed but no workflow ran them; a broken helper would have burned all fifteen jobs before anyone saw why.

## Receipts

`bash scripts/first-run/lib.test.sh` → `PASS: lib.test.sh`, rc=0, including eight new assertions:

```
  ok   check_eventually PASS on the first try
  ok   check_eventually reports no wait
  ok   check_eventually PASS once state appears
  ok   check_eventually reports the wait
  ok   check_eventually did wait for it
  ok   check_eventually FAIL after the cap
  ok   check_eventually FAIL names the cap
  ok   check_eventually records one row
```

Mutation control — swap the delayed case back to a plain immediate `check` (`sed -i 's/^check_eventually ce-later 20 -- test/check ce-later -- test/'`) and the suite fails, rc=1:

```
  FAIL check_eventually PASS once state appears
  FAIL check_eventually reports the wait
FAIL: lib.test.sh (2 assertion(s))
```

Reverted, rc=0 again.

`shellcheck scripts/first-run/lib.sh scripts/first-run/macos-app.sh scripts/first-run/lib.test.sh` rc=0; `bash -n` on all three rc=0; `actionlint .github/workflows/first-run-gauntlet.yml` rc=0.

## Follow-up not in this PR

`lib.test.sh` still does not gate pull requests — only gauntlet dispatches. Wiring it into `ci.yml` means a new filter output, a new job, and matching `needs` and `expect_job` entries in the `conclusion` aggregator, which is the file that can block every PR if the skip/run parity is wrong. That belongs in its own change.

### DIFF
--- a/.github/workflows/first-run-gauntlet.yml
+++ b/.github/workflows/first-run-gauntlet.yml
@@ -77,6 +77,12 @@ jobs:
     steps:
       - uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6
 
+      # Every channel script sources lib.sh, so a broken helper would burn all
+      # fifteen jobs below before anyone saw why. Seconds here, and the only
+      # place these unit tests run.
+      - name: Self-test the gauntlet helper library
+        run: bash scripts/first-run/lib.test.sh
+
       # The release JSON is fetched once and reused by the inventory and parity
       # steps below so every job in the run talks about the same asset list.
       - name: Resolve release tag
--- a/scripts/first-run/lib.sh
+++ b/scripts/first-run/lib.sh
@@ -13,6 +13,8 @@
 #   check NAME -- CMD ARGS...            PASS when CMD exits 0
 #   check_output NAME SUBSTR -- CMD...   PASS when CMD exits 0 and output contains SUBSTR
 #   check_fails NAME SUBSTR -- CMD...    PASS when CMD exits nonzero and output contains SUBSTR
+#   check_eventually NAME SECS -- CMD...  PASS as soon as CMD exits 0, retrying
+#                                        once a second for up to SECS; one row
 #   info NAME VALUE                      informational row (never fails the run)
 #   wait_health URL SECS                 poll /api/health; records seconds-to-health; returns 1 on timeout
 #   assert_version URL EXPECTED          PASS when health .version == EXPECTED (leading v stripped)
@@ -106,6 +108,29 @@ check_fails() {
         _gauntlet_record PASS "$name" "$rc" "$out"
     else
         _gauntlet_record FAIL "$name" "$rc" "expected nonzero exit with substring: $want; got: $out"
+    fi
+    return 0
+}
+
+# For state some other process writes shortly after the signal we waited on —
+# a launch agent plist, a spawned process — where testing once immediately
+# races the writer. Records a single row: the elapsed seconds on PASS, the last
+# attempt's output on FAIL.
+check_eventually() {
+    local name="$1" secs="$2"; shift 2
+    [ "${1:-}" = "--" ] && shift
+    local i rc=1
+    for ((i = 1; i <= secs; i++)); do
+        rc="$(_gauntlet_run "$name" "$@")"
+        [ "$rc" = 0 ] && break
+        [ "$i" -lt "$secs" ] && sleep 1
+    done
+    local out
+    out="$(cat "$GAUNTLET_OUT/checks/$name.log" 2>/dev/null)"
+    if [ "$rc" = 0 ]; then
+        _gauntlet_record PASS "$name" "$rc" "ready after $((i - 1))s${out:+; $out}"
+    else
+        _gauntlet_record FAIL "$name" "$rc" "still failing after ${secs}s${out:+; $out}"
     fi
     return 0
 }
--- a/scripts/first-run/lib.sh
+++ b/scripts/first-run/lib.sh
@@ -13,8 +13,9 @@
 #   check NAME -- CMD ARGS...            PASS when CMD exits 0
 #   check_output NAME SUBSTR -- CMD...   PASS when CMD exits 0 and output contains SUBSTR
 #   check_fails NAME SUBSTR -- CMD...    PASS when CMD exits nonzero and output contains SUBSTR
-#   check_eventually NAME SECS -- CMD...  PASS as soon as CMD exits 0, retrying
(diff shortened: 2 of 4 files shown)


7xuanlu:
## Codex Sol adversarial pass (gpt-5.6-sol, one round) — DO NOT MERGE on `8a755bd`, fixed in `6716ace`

| # | Finding | Outcome |
|---|---|---|
| P1 | Waiting for the plist file can pass a *failed* install: `install_app_plist` (`app/src/lifecycle.rs:299-333`) writes the file, runs `launchctl load`, and deletes the file again on load failure, so the poll can catch it mid-window. | Confirmed by reading the code — the repo already documents the same hazard for the server plist at `lifecycle.rs:2093-2096`. Fixed by waiting for the **loaded launchd job** instead: that state is only reachable after a load that stuck. Adds `launchctl-desktop-loaded`, which had no equivalent before, and the file checks now run on a settled install. The headless fallback probes the launchd *domain* (`gui/$UID`) rather than a label, so "not loaded yet" cannot read as "no launchd here". |
| P2 | `check_eventually` counted `SECS` attempts but only `SECS-1` sleeps, so a 2 s cap took 1 s while reporting 2; each attempt got the full `GAUNTLET_TIMEOUT`; `SECS=0` ran nothing and could kill a caller under `set -e` reading a missing log; non-numeric `SECS` aborted under `set -u`. | Fixed: one wall-clock deadline, each attempt handed the time still left as its `GAUNTLET_TIMEOUT`, a `FAIL` row naming a bad `SECS`, and `|| true` on the log read. Measured: `check_eventually never-5 5 -- false` takes 5 s and records `still failing after 5s`. The per-attempt cap is only enforced where `timeout`/`gtimeout` exists, same as every other check here — now stated in the comment rather than implied. |
| P2 | The new test coordinated on wall-clock time (a 2 s background writer), so a descheduled runner could flip it — and it now gates every gauntlet dispatch. | Fixed: the probe fails until it has been called three times, so the retry assertions count attempts and depend on nothing timing-related. |
| P3 | `plist-server-exists` does not have the diagnosed race, so a 30 s retry only delays a genuine failure. | Taken: the loaded-job check runs first, so both file checks now run on a settled install and return on the first try. |

**Found while proving the P2 fix, and fixed here too:** `lib.test.sh` could abort mid-run and still exit 0. A bare `rm -rf` EXIT trap overwrites the status, and in bash 3.2 (the macOS system shell) `$?` inside the trap does not carry a `set -u` abort either. The suite now passes only when it reaches its last line. This matters because the same commit makes the suite gate every gauntlet dispatch.

### Receipts on `6716ace`

`bash scripts/first-run/lib.test.sh` → `PASS: lib.test.sh`, rc=0, twelve `check_eventually` assertions.

Mutation controls, each run by hand and reverted:

- Retry removed (`[ "$rc" = 0 ] && break` → `break`): 3 assertions fail, rc=1 — `PASS once state appears`, `retries until it passes`, `keeps the last output`.
- `SECS` validation removed: the suite aborts with `lib.sh: line 133: x: unbound variable` and now exits 1. Before the trap fix, that same abort exited **0** — which is how the gap was found.

`shellcheck` on all three scripts rc=0; `bash -n` on all three rc=0; `actionlint .github/workflows/first-run-gauntlet.yml` rc=0.

### Still open

`launchctl-desktop-loaded` is new and unproven on a runner — the previous gauntlet runs never printed that label. Evidence that the domain works there: run 33124875152 recorded `launchctl-server-loaded PASS` with `gui/501/com.wenlan.server ... state = running`. If the desktop job legitimately does not load on a CI runner this check goes red, which is itself worth knowing about run-at-login.

PR #621, opened 2026-08-28
fix(deps): raise five transitive npm packages past their advisories

7xuanlu:
`pnpm audit` on `main` reports **10 advisories — 5 high, 5 moderate**, every one of them in a package nothing in this repository depends on directly. GitHub's Dependabot alerts, switched on today, report 7 of the same set.

| Package | Was | Now | Advisories |
|---|---|---|---|
| brace-expansion | 5.0.7 | 5.0.9 | GHSA-mh99-v99m-4gvg, GHSA-rgw5-rvv9-x895 (both high) |
| undici | 7.28.0 | 7.29.0 | GHSA-4cwx-7wf7-3272 (high), GHSA-8xcm-r25x-g524, GHSA-jr45-8vmc-qm54, GHSA-m8rv-5g2x-5cg5, GHSA-v3r7-h72x-cjcm |
| fast-uri | 3.1.4 | 3.1.6 | GHSA-7p8r-x3mc-p8w7 (high) |
| nanoid | 3.3.16 | 3.3.18 | GHSA-2v37-7h3g-55p8 (high) |
| postcss | 8.5.20 | 8.5.26 | GHSA-fxqj-rqcc-2cmp |

## Why overrides and not an update

`pnpm update brace-expansion postcss undici fast-uri nanoid --depth Infinity` fixes all five, but re-resolves far beyond them: 334 lockfile lines moved, pulling new Sentry, OpenTelemetry, Babel, ESLint and Vite-plugin versions along for the ride. That is a large surface to review for five security bumps.

Each override here is written as `"<pkg>@<vulnerable range>": "^<smallest safe version>"`, which does two things worth keeping:

- **Scoped to the vulnerable range**, so a copy of the package that already resolves to a safe version is left exactly where it is.
- **Held inside the major the parents already use.** Writing `">=3.1.5"` instead of `"^3.1.5"` is what a plain minimum looks like, and it resolved fast-uri to 4.1.3, nanoid to **6.0.1** (from 3.x, where 4.0 went ESM-only) and undici to 8.10.0 — three unrequested major bumps. The caret form keeps them on 3.1.6, 3.3.18 and 7.29.0.

The lockfile moves 49 lines.

The `comment` key next to `overrides` says when an entry can go: once every parent resolves past it on its own, `pnpm audit` stays clean without it.

## Receipts

- `pnpm audit` → **No known vulnerabilities found** (was 10)
- `pnpm install --frozen-lockfile` → clean, so `package.json` and the lockfile agree
- `pnpm exec tsc --noEmit -p tsconfig.json` → `TypeScript: No errors found`, rc=0
- `pnpm build` (`tsc -b && vite build`) → rc=0, built in 2.74s — this is the path postcss is actually on
- `pnpm exec vitest run` → **1979 passed**, 2 failed, 26 skipped. The two failures are the known environment ones in `scripts/dev-runtime.test.ts` (`isolated dev port 27991 is already in use`, and 27992), caused by another worktree's dev runtime holding those ports on this machine. Same two, same count as the recorded baseline.

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,7 +101,7 @@
     "comment": "overrides: smallest safe version for an advisory, held inside the major the parents already use, and applied only to the vulnerable range so an already-safe copy is left alone. Drop an entry once every parent resolves past it by itself; `pnpm audit` is the check.",
     "overrides": {
       "brace-expansion@>=4.0.0 <5.0.9": "^5.0.9",
-      "fast-uri@>=3.0.0 <3.1.5": "^3.1.5",
+      "fast-uri@>=3.0.0 <3.1.6": "^3.1.6",
       "nanoid@<3.3.18": "^3.3.18",
       "postcss@<=8.5.22": "^8.5.23",
       "undici@>=7.0.0 <7.29.0": "^7.29.0"
--- a/package.json
+++ b/package.json
@@ -97,7 +97,15 @@
   "pnpm": {
     "onlyBuiltDependencies": [
       "esbuild"
-    ]
+    ],
+    "comment": "overrides: smallest safe version for an advisory, held inside the major the parents already use, and applied only to the vulnerable range so an already-safe copy is left alone. Drop an entry once every parent resolves past it by itself; `pnpm audit` is the check.",
+    "overrides": {
+      "brace-expansion@>=4.0.0 <5.0.9": "^5.0.9",
+      "fast-uri@>=3.0.0 <3.1.5": "^3.1.5",
+      "nanoid@<3.3.18": "^3.3.18",
+      "postcss@<=8.5.22": "^8.5.23",
+      "undici@>=7.0.0 <7.29.0": "^7.29.0"
+    }
   },
   "repository": {
     "type": "git",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -6,7 +6,7 @@ settings:
 
 overrides:
   brace-expansion@>=4.0.0 <5.0.9: ^5.0.9
-  fast-uri@>=3.0.0 <3.1.5: ^3.1.5
+  fast-uri@>=3.0.0 <3.1.6: ^3.1.6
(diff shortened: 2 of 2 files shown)


7xuanlu:
Adversarial review done (Codex Sol, gpt-5.6-sol, xhigh). Verdict: **MERGE**, with one P3 acted on in `e3e8397`.

**P3 — the fast-uri override floor was one patch short.** Four high-severity advisories published 2026-08-23 affect the 3.x line through 3.1.5 and are fixed in 3.1.6. The override was keyed on `>=3.0.0 <3.1.5`, so a dependency pinned at exactly 3.1.5 would have been left alone. Raised to `>=3.0.0 <3.1.6: ^3.1.6`. The lockfile already resolved 3.1.6, so nothing moved — `pnpm install --lockfile-only` changed only the two override lines.

**What the review cleared:**

- Every target resolves to exactly one package and snapshot entry: `brace-expansion 5.0.9`, `fast-uri 3.1.6`, `nanoid 3.3.18`, `postcss 8.5.26`, `undici 7.29.0`. No vulnerable version or peer-resolution variant survives anywhere in the tree.
- All parent constraints accept the resolved versions, so no override forces a version a parent rejects.
- Manifest overrides match the lockfile exactly: all 29 production and 21 development importer specifiers, no extras, no mismatches. `git diff --check` returned 0.
- The `"comment"` key inside the `pnpm` object is ignored safely by pnpm 10.28.2. The root package is private, and release publishing runs from the two separate `crates/*/npm` directories, so the key is neither rejected nor published.
- The five packages are reachable only through development, build, and test tooling. No repository source imports any of them.
- The only other lockfile, `scripts/readme-visuals/package-lock.json`, holds Playwright and none of the five.

**One thing stayed unchecked:** live `pnpm audit` could not run in the reviewer's read-only environment — registry DNS was unavailable (`ENOTFOUND`, rc=1). The committed versions and advisory ranges were verified from the lockfile and the published advisories instead.

PR #623, opened 2026-08-28
fix(security): close the three remaining audit findings on the open and relay paths

7xuanlu:
Closes the three remaining security findings from the 2026-08-16 audit. All three were still live in `main`; each was re-checked against the shipped tree before any of this was written.

## S5 — the app would open, or run, whatever a memory pointed at

`open_file` in `app/src/search.rs` handed its argument straight to the desktop's default handler. That argument is not the app's own: `POST /api/ingest/webpage` and `/api/ingest/memory` store the caller's `url` verbatim (`crates/wenlan-server/src/ingest_routes.rs`), it rides out on every search result, `searchResultTarget` in `src/lib/searchResultTarget.ts` turns a result with a `url` into a `file` target, and `openSearchResult` in `src/components/memory/Main.tsx` opened it. So any agent or page that can put a memory in the store chose what the OS would act on.

The app's other opener, the Tauri shell plugin behind citation links, has been scheme-restricted all along by the plugin's default scope. This one was not.

### The fix: two commands, split by who controls the input

One command served two callers with two different trust levels. The Sources view passes a path the user picked in a folder they chose. A search result passes a `url` an attacker may control. A single check had to stay loose enough for the first caller, and that slack is what the second caller could use.

**`open_file`** keeps the user-chosen paths and a denylist. A URL must carry `http` or `https` — that closes `javascript:`, `data:`, `vbscript:`, and the long tail of registered application schemes. A bare path must not end in a launcher extension for this platform (`app`, `command`, `pkg`, `scpt`, `workflow`, `webloc` on macOS; the `exe`/`bat`/`ps1`/`lnk` set on Windows; `desktop` elsewhere) and, on Unix, must not be a regular file with an executable bit. Directories are exempt because the Sources view opens folders in the file manager and every folder carries that bit. A one-character scheme is read as a Windows drive letter, so `C:\Users\...` still opens.

**`open_search_result`** is new and takes the attacker-influenced input. It answers a narrower question — is this a web link, or a real file of a type the daemon indexes? — with an allowlist:

- A URL must carry `http` or `https`.
- A path is resolved with `std::fs::canonicalize` **before** it is judged. A symlink named `notes.md` pointing at an `.app` bundle is seen for what it resolves to, and `..` cannot walk anywhere the check did not look.
- It must resolve to a regular file. `.app` and `.pkg` bundles are directories, so this alone stops both.
- Its extension must be `md`, `txt`, or `pdf` — the three `SUPPORTED_EXTENSIONS` in `src/components/memory/SourcesView.tsx` says the daemon indexes.

The OS is handed the canonical path the check ran against, not the string it arrived as. Re-resolving the original would leave a window in which a symlink swapped after the check and before the open sends the OS somewhere the check never saw.

Both commands refuse an embedded NUL byte. On Windows the `open` crate hands `ShellExecuteExW` a UTF-16 buffer whose encoder preserves interior NULs (`windows.rs:350`), and Win32 truncates at the first one — so `report.pdf\0.exe` would have been checked as a PDF and run as a program.

Both frontends now surface a refusal instead of failing silently: `Main.tsx` through `toast.error`, `SourcesView.tsx` through the toast idiom already in that file.

## S2 — the relay ID was guessable

The Remote Access relay ID was `u` plus 11 hex characters of a `DefaultHasher` over the wall clock and the process ID: a hash with fixed keys, over two guessable inputs, truncated to 44 bits.

That ID is the whole of the protection on the remote endpoint. It is the path segment of the URL handed to Claude.ai and ChatGPT, and `register_with_relay` posts the same value as `secret`.

It now comes from `getrandom`, 16 bytes rendered as 32 hex characters, matching the pattern already used in `app/src/presence.rs`. An ID in the old shape is replaced rather than reused, since keeping one would leave the endpoint as reachable as it was.

## S6 — a bearer token that nothing checked

The app generated a 0600 token file on the connect path, could fail the whole connect if that generation failed, carried the token on `RemoteAccessStatus::Connected`, and exposed a `rotate_remote_token` command — while `spawn_mcp`, the only place the app starts `wenlan-mcp serve`, always passes `--no-auth`. With `--no-auth` the sidecar installs no auth middleware, so nothing ever read the token. The local Claude Desktop path does not use it either: that one runs over stdio via `npx wenlan-mcp`. No part of the UI showed or rotated it, and the frontend wrapper had no production caller.

The machinery implied a protection that was not wired, which is the finding. It is deleted, which also drops a written secret, a subprocess call, and a failure mode from the connect path. A comment at the connect site now says why there is no token: the Claude.ai and ChatGPT connectors cannot attach an `Authorization` header, so `--no-auth` is deliberate, and what guards the endpoint is the unguessable relay ID from S2, the explicit in-app warning, and the feature being off by default.

`crates/wenlan-mcp`'s own token support is untouched — that binary still enforces a token when it is run with one, and the app does not delete an existing `~/.config/wenlan-mcp/token` that a manually configured server may still rely on.

## Verification

Every fix that adds behavior has a RED control. Each was reverted against the committed tree and re-run, not assumed:

| Break | Result |
|---|---|
| `refuse_unopenable_target` always returns `Ok(())` (`git diff --stat`: 2 insertions, 6 deletions in `app/src/search.rs`) | `test result: FAILED. 2 passed; 2 failed` — `a_memory_url_cannot_reach_another_scheme`, `a_refusal_says_which_scheme_and_what_is_allowed` |
| the `open_file` path gate short-circuited to never fire (`git diff --stat`: 1 insertion, 1 deletion) | `test result: FAILED. 4 passed; 3 failed` — `a_memory_path_cannot_launch_an_application`, `an_executable_file_is_refused_and_a_document_is_not`, `a_refused_program_says_which_one` |
| `refuse_unopenable_search_result` returns `Ok(None)` for everything, i.e. the allowlist removed (`git diff --stat`: 2 insertions, 9 deletions) | `test result: FAILED. 13 passed; 7 failed` — the scheme, missing-file, directory, executable-script, Windows-executable, and both symlink cases |
| canonicalization dropped, the original path judged instead (`git diff --stat`: 1 insertion, 2 deletions) | `test result: FAILED. 19 passed; 1 failed` — `a_search_result_symlink_disguised_as_markdown_is_refused` |
| `relay_id_is_current` reduced to `!id.is_empty()` (`git diff --stat`: 1 insertion, 5 deletions in `app/src/remote_access.rs`) | `test result: FAILED. 6 passed; 2 failed` — `relay_id_shape_rejects_the_old_and_the_malformed`, `a_relay_id_from_before_the_csprng_is_replaced` |

Full receipts on the final tree (`f1d4d24c`):

```
cargo fmt --all                                          rc=0
cargo clippy -p wenlan-app --lib --tests -- -D warnings   rc=0
cargo test -p wenlan-app --lib                           rc=0
  test result: ok. 465 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out
npx tsc -b --force                                       rc=0
vitest run (6 affected files)                            rc=0
  Test Files  6 passed (6) / Tests  120 passed (120)
```

Exit codes were captured at each command rather than read off a pipeline tail.

## Note for reviewers

Existing installs with Remote Access already set up get a new relay URL on next start, because the old relay ID is replaced. That is intended — the old ID is the guessable one. A saved connector has to be re-added. The feature is off by default and marked experimental.

### DIFF
--- a/app/src/lib.rs
+++ b/app/src/lib.rs
@@ -1307,6 +1307,7 @@ pub fn run() {
             search::delete_by_time_range,
             search::delete_bulk,
             search::open_file,
+            search::open_search_result,
             search::read_source_dir,
             search::detect_obsidian_vaults,
             search::read_text_file,
--- a/app/src/lib.rs
+++ b/app/src/lib.rs
@@ -1423,7 +1423,6 @@ pub fn run() {
             // Remote access commands
             search::toggle_remote_access,
             search::get_remote_access_status,
-            search::rotate_remote_token,
             search::test_remote_mcp_connection,
             // Memory nurture commands
             search::get_nurture_cards_cmd,
--- a/app/src/remote_access.rs
+++ b/app/src/remote_access.rs
@@ -197,7 +197,6 @@ pub enum RemoteAccessStatus {
     Starting,
     Connected {
         tunnel_url: String,
-        token: String,
         /// Stable relay URL (if relay registration succeeded).
         relay_url: Option<String>,
     },
@@ -775,80 +774,12 @@ fn write_private_file(path: &Path, contents: &[u8], file_name: &str) -> Result<(
     restrict_private_file(path, file_name)
 }
 
-fn import_nonempty_legacy_file(
-    current_dir: &Path,
-    legacy_dir: &Path,
-    file_name: &str,
-) -> Result<PathBuf, String> {
-    let current_path = current_dir.join(file_name);
-    if current_path.exists() {
-        restrict_private_dir(current_dir, file_name)?;
-        restrict_private_file(&current_path, file_name)?;
-        return Ok(current_path);
-    }
-
-    let legacy_path = legacy_dir.join(file_name);
-    let contents = match std::fs::read_to_string(&legacy_path) {
-        Ok(contents) => contents,
-        Err(e) if e.kind() == std::io::ErrorKind::NotFound => return Ok(current_path),
-        Err(e) => {
-            log::warn!(
-                "[remote-access] Skipping legacy {} import from {}: {}",
-                file_name,
-                legacy_path.display(),
-                e
-            );
-            return Ok(current_path);
-        }
-    };
-
-    if contents.trim().is_empty() {
-        return Ok(current_path);
-    }
-
-    write_private_file(&current_path, contents.as_bytes(), file_name)?;
-
-    Ok(current_path)
-}
-
-fn token_file_path_for_dirs(current_dir: &Path, legacy_dir: &Path) -> Result<PathBuf, String> {
-    import_nonempty_legacy_file(current_dir, legacy_dir, "token")
-}
-
 fn relay_id_path_for_dirs(current_dir: &Path) -> PathBuf {
     current_dir.join("relay_id")
 }
 
-/// Token file path for wenlan-mcp authentication.
-/// wenlan-mcp stores tokens at ~/.config/wenlan-mcp/token (XDG convention),
-/// NOT ~/Library/Application Support/ (macOS convention from dirs::config_dir).
-fn token_file_path() -> Result<PathBuf, String> {
-    let current = crate::identity_paths::mcp_config_dir();
-    if crate::identity_paths::isolated_dev_state_dir().is_some() {
-        return token_file_path_for_dirs(&current, &current);
-    }
-    token_file_path_for_dirs(&current, &crate::identity_paths::legacy_mcp_config_dir())
-}
-
 fn relay_id_path() -> PathBuf {
     relay_id_path_for_dirs(&crate::identity_paths::mcp_config_dir())
-}
-
-fn token_generate_args(path: &std::path::Path) -> Vec<String> {
-    vec![
-        "token".to_string(),
-        "generate".to_string(),
-        "--output".to_string(),
-        path.to_string_lossy().into_owned(),
-    ]
-}
-
-/// Read the bearer token from disk.
-pub fn read_token() -> Result<String, String> {
-    let path = token_file_path()?;
-    std::fs::read_to_string(&path)
-        .map(|s| s.trim().to_string())
-        .map_err(|e| format!("Failed to read token at {}: {}", path.display(), e))
 }
 
 /// Start the remote access tunnel (wenlan-mcp serve + cloudflared).
@@ -907,7 +838,7 @@ async fn toggle_on_inner(
     let result = start_tunnel(&app_handle, operation_generation).await;
 
     match result {
-        Ok((tunnel_url, token, mcp_child, tunnel_child, tunnel_owner, port, mcp_rx, tunnel_rx)) => {
+        Ok((tunnel_url, mcp_child, tunnel_child, tunnel_owner, port, mcp_rx, tunnel_rx)) => {
             {
                 let state = app_handle
                     .state::<std::sync::Arc<tokio::sync::RwLock<crate::state::AppState>>>();
@@ -934,7 +865,6 @@ async fn toggle_on_inner(
 
             let status = RemoteAccessStatus::Connected {
                 tunnel_url: tunnel_url.clone(),
-                token: token.clone(),
                 relay_url: relay_url.clone(),
             };
             // Store process handles in state
@@ -1567,7 +1497,6 @@ async fn start_tunnel(
 ) -> Result<
     (
         String,                                                                 // tunnel_url
-        String,                                                                 // token
         tauri_plugin_shell::process::CommandChild,                              // mcp_child
         tauri_plugin_shell::process::CommandChild,                              // tunnel_child
         CloudflaredOwner, // tunnel owner receipt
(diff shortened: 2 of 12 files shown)


7xuanlu:
## Round 2 — all four findings acted on, at `f1d4d24c`

The review's core point was that the denylist was the wrong shape for the input, not that it had holes to patch. It is right, and the fix follows the durable path it named: **separate the trusted Sources operation from search-result URLs.**

### P1 (NUL truncation) — fixed

Both commands now refuse a target containing `\0` before anything else. Confirmed the mechanism in the pinned crate: the encoder at `open-5.4.1/src/windows.rs:350` preserves interior NULs in the UTF-16 buffer it hands `ShellExecuteExW` at `windows.rs:291`, and Win32 truncates there. `report.pdf\0.exe` would have been checked as a PDF and run as a program.

### P1 (`.appref-ms`, and the symlink-to-a-directory bypass) — fixed by the split, not by extending the list

Adding `.appref-ms` to a denylist would have left the next omission open, and the macOS symlink case is not an extension problem at all. So `open_search_result` is now a separate command with an **allowlist**:

- `http`/`https`, or
- a path that `std::fs::canonicalize` resolves to a **regular file** whose extension is `md`, `txt`, or `pdf`.

That kills both worked examples at once. `Quarterly Notes -> Terminal.app` canonicalizes to the bundle, which is a directory, and a directory is no longer openable through this command at all. `invoice.appref-ms` is not one of the three indexed types. `.appref-ms` was added to the `open_file` denylist as well, for the trusted caller.

The OS is handed the canonical path the check ran against, not the original string — otherwise a symlink swapped between check and open would still win.

### P2 (0755 Markdown refused into a void) — fixed on both halves

The false positive is gone by construction: `open_search_result` never consults the executable bit, so a 0755 `README.md` opens. `open_file` keeps the bit check for the trusted caller, where it costs nothing real.

Both call sites now show the refusal. `Main.tsx:761` catches and raises `toast.error`; `SourcesView.tsx:648` uses the `toast(..., { description })` idiom already in that file.

### P3 (test mocks kept the deleted `token`) — fixed

Removed from all three sites and each response object now carries `satisfies RemoteAccessStatus`, so a future drift in the serialized contract fails `tsc` instead of passing silently.

### Verification

Two new RED controls, run against the committed tree:

| Break | Result |
|---|---|
| `refuse_unopenable_search_result` returns `Ok(None)` for everything (allowlist removed) — `git diff --stat`: 2 insertions, 9 deletions | `test result: FAILED. 13 passed; 7 failed` — scheme, missing file, directory, executable script, Windows executable, and both symlink cases |
| canonicalization dropped, original path judged — `git diff --stat`: 1 insertion, 2 deletions | `test result: FAILED. 19 passed; 1 failed` — `a_search_result_symlink_disguised_as_markdown_is_refused` |

The three earlier controls still hold. Receipts on `f1d4d24c`: clippy `rc=0`, `cargo test -p wenlan-app --lib` `rc=0` with 465 passed / 0 failed, `npx tsc -b --force` `rc=0`, vitest 6 files / 120 tests passed. The two symlink tests build the symlink on disk in a temp dir rather than asserting on a string.

### One point not taken

The review's alternative of a reveal-only operation (select the file in the file manager instead of opening it) would be safer still, but it changes what the feature does — clicking a search result would stop opening the document. The allowlist keeps the behavior and confines it to the three types the daemon actually indexes. Worth revisiting if that set ever widens to a format with an active-content history.

7xuanlu:
## Round 3 — CI found a regression the round-2 tests did not, fixed at `c9abd1ba`

`app-check` failed one frontend test. The stale assertion was the visible edge of a real bug, so this is worth stating plainly rather than filing as a test fix.

`app/src/sources/local_files.rs:193` stores every local-file source's `url` as `format!("file://{}", path.display())`. So a real search result target is a `file://` URL, not a bare path. The old path invoked `open_file`, whose TypeScript wrapper strips that prefix before invoking, so Rust saw a bare path. The new `open_search_result` did not strip it and allows only `http` and `https` — which would have refused **every source file in the app**. Round 2 shipped a working guard attached to the wrong input shape.

### The fix

The prefix is stripped inside `refuse_unopenable_search_result`, before the scheme check, not in the frontend wrapper. The guard has to be complete on its own: a caller that forgets to strip must get the same answer as one that remembers.

Nothing else about the decision changes. What remains still goes through canonicalization, the regular-file check, and the indexed-extension check, so `file:///Applications/Utilities/Terminal.app` is still refused — it canonicalizes to a directory, and there is a test for exactly that. No percent-decoding was added: the producer applies none, so accepting encodings it never emits would widen the input surface for nothing.

### Verification

RED control: removing only the two strip lines turns `a_file_url_naming_a_real_document_is_allowed` red with the `file:` refusal message — 21 passed, 1 failed — and `git diff --stat` shows 2 deletions and nothing else on the tree.

Receipts on the integrated commit, after merging `main` in (a clean merge proves the text reconciled, not that the meaning survived, so this is a fresh run rather than the pre-merge one):

```
cargo clippy -p wenlan-app --lib --tests -- -D warnings   rc=0
cargo test -p wenlan-app --lib                           rc=0
  test result: ok. 467 passed; 0 failed; 1 ignored
npx tsc -b --force                                       rc=0
vitest run (full suite)                                  rc=0
  Test Files  162 passed | 1 skipped (163)
  Tests  1955 passed | 2 skipped (1957)
```

`scripts/dev-runtime.test.ts` is excluded from that local run only: two of its cases fail on this machine because dev port 27992 is held by another worktree's daemon (`error: isolated dev port 27992 is already in use`). They pass in CI, which starts clean. Nothing in this branch touches that script.

### What I took from it

The round-2 local run covered the six files the diff touched. This test lives in a seventh that exercises the same code, and only the full suite would have caught it. The full frontend suite ran before this push.

Issue #617, opened 2026-08-27
deps: PDF text extractor must reach lopdf 0.42 without regressing font decoding (RUSTSEC-2026-0187)


7xuanlu:
## Observed

A one-time `cargo audit` over main (`9e3a8bee`, RustSec advisory DB as of 2026-08-27) reports 11 vulnerabilities. Four are fixed by compatible `cargo update` bumps (crossbeam-epoch, h2, quinn-proto, rustls-webpki; landing in the pre-launch trust PR). The rest need a major-version upgrade of a direct or transitive dependency:

| Advisory | Crate | Fix | Why it matters |
|---|---|---|---|
| RUSTSEC-2026-0187 | `lopdf` 0.38.0 | ≥0.42.0 | stack overflow on deeply nested PDF objects: a crafted PDF source can crash the daemon during ingest (CVSS 7.5, availability) |
| RUSTSEC-2026-0195, RUSTSEC-2026-0194 | `quick-xml` 0.30.0 and 0.39.4 (two copies in the graph) | ≥0.41.0 | memory exhaustion / quadratic parsing on attacker-controlled XML (CVSS 7.5) |

## To do

- Find the dependents: `cargo tree -i lopdf`, `cargo tree -i quick-xml@0.30.0`, `cargo tree -i quick-xml@0.39.4`.
- Upgrade `lopdf` to 0.42 (API changes in `Document`/`Object` access) and re-run the PDF ingest tests.
- Move the quick-xml dependents to a version on ≥0.41, or replace the crate that pins 0.30.
- Re-run `cargo audit` and record the result in the audit tracker.

Also flagged as warnings only (not blocking): `unic-ucd-version` unmaintained, unsound notes on `anyhow` 1.0.102, `glib` 0.18.5, `lru` 0.12.5, `rand` 0.8.5/0.9.2.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01QEPWkx6eM4zNpDG9VB2nZh

7xuanlu:
## Status after PR #625: two of three closed, the PDF one held on purpose

[...]

PR #625, opened 2026-08-28
fix(deps): raise plist past the quick-xml advisories

7xuanlu:
Fixes two of the three RustSec advisories tracked in issue #617. The third is deliberately held, with evidence below. **Issue #617 stays open** — do not auto-close it on merge.

## What ships

`plist` 1.9.0 → 1.10.0, which moves `quick-xml` 0.39.4 → 0.41.0.

| Advisory | Fixed in `quick-xml` 0.41.0 by |
|---|---|
| RUSTSEC-2026-0195 — memory exhaustion on hostile XML | a namespace limit of 256 (`quick-xml-0.41.0/src/name.rs:522`) |
| RUSTSEC-2026-0194 — quadratic duplicate-attribute scan | switching large attribute sets to hashing (`quick-xml-0.41.0/src/events/attributes.rs:969`) |

`plist` is reached through `service-manager` (the launchd plists the CLI and server write) and through Tauri itself. `quick-xml` 0.41.0 was already in the graph, so this **removes** a duplicate version rather than adding one.

The diff is `Cargo.lock` only. No manifest moves, because `plist` is transitive.

## What is held, and why

**RUSTSEC-2026-0187** — `lopdf` 0.38.0 overflows the stack on deeply nested PDF objects — stays open. The only route to the fixed `lopdf` 0.42 is `pdf-extract` 0.12.0, and 0.12.0 silently corrupts text extraction.

Upstream commit [`205c16ff`](https://github.com/jrmuizel/pdf-extract/commit/205c16ff) ("Avoid reparsing fonts for every page") moved the font cache out of `process_stream` and onto `Processor`, which lives for the whole document. It is keyed on the local resource name alone:

```rust
// pdf-extract 0.12.0, src/lib.rs:1711 — cache is a Processor field, document-lived
let font = self.font_table.entry(name.to_owned()).or_insert_with(...)

// pdf-extract 0.10.0, src/lib.rs:1709 — cache is a local, rebuilt per stream
let font = font_table.entry(name.to_owned()).or_insert_with(...)
```

A PDF's `/F1` is scoped to the page or form that names it. Once 0.12 caches one, every later `/F1` inherits the first font's encoding and Unicode map no matter which font object it actually points at. The same `Processor` also recurses into forms carrying their own resource dictionaries.

This is not hypothetical for us. An adversarial review counted **75** conflicting font-object uses in our own ResNet acceptance fixture and **443** in the BERT one; in ResNet, `/F1` moves between object 57 (`Times-Roman`), a custom Type1 face, and `CMEX10`.

Both callers of `pdf_extract::extract_text_from_mem` — `crates/wenlan-core/src/sources/directory.rs:401` and `app/src/sources/local_files.rs:210` — receive this as `Ok(String)`. Nothing detects it before the text is normalized, indexed, and stored.

Trading a crash advisory for silently wrong memories is the wrong direction. A crash is visible; a mis-decoded page is recalled later as fact.

`crates/wenlan-core/tests/pdf_ordered_text.rs` passes on 0.12 without catching this, which is worth saying plainly: it checks four hand-picked abstract and introduction pairs plus a 500-word floor. That makes it a reading-order gate, not a font-decoding one.

There is no version between 0.10.0 and 0.12.0 to fall back to, and `205c16ff` is still the newest touch to that code upstream.

A local fix through `[patch.crates-io]` **is** possible. An earlier draft of this PR claimed it would block publishing `wenlan-core`; that was wrong, and the review caught it. Cargo strips the whole patch table when it packages a crate — verified here by packaging a scratch crate carrying one and reading the manifest back out of the `.crate` archive, which contains no `[patch]` section.

What it would actually cost: the patch does not travel with the published crate, so anyone consuming `wenlan-core` from crates.io still resolves the corrupting `pdf-extract`, and we would be maintaining a fork of a PDF parser. That is a standing commitment rather than a one-off fix, which is why it is not being taken unilaterally here.

## Verification

Run with the Bash sandbox disabled — `~/.cargo/registry/src/` is not writable inside it, so cargo cannot unpack a crate version it has never seen and reports the failure as `rc=101`, which reads exactly like a compile error.

```
cargo clippy --workspace --exclude wenlan-app --all-targets -- -D warnings   rc=0
cargo clippy -p wenlan-app --lib --tests -- -D warnings                      rc=0
cargo test -p wenlan-app --lib     rc=0     447 passed, 0 failed
cargo test -p wenlan-core --lib    rc=0    4162 passed, 0 failed
cargo test -p wenlan --lib         rc=0      59 passed, 0 failed
```

`cargo tree -i quick-xml@0.39.4 --workspace --target all` no longer matches any package. `cargo tree -i plist` reports 1.10.0.

## Also confirmed, not changed

`quick-xml` 0.30.0 stays in the graph and the advisories do not reach it. It is a **build**-dependency of `xcb` (`xcb-1.7.0/Cargo.toml:280`), made Linux-only by `xcap` (`xcap-0.9.0/Cargo.toml:150`), and its build script parses only `.xml` files under the crate's own directory (`xcb-1.7.0/build/main.rs:26`). It uses plain `Reader`, so RUSTSEC-2026-0195's `NsReader` path is absent entirely.

Exactly one `sysinfo` package remains in `Cargo.lock`, as `AGENTS.md` requires.


7xuanlu:
## Correction: `[patch.crates-io]` does not block publishing

The confirmation review cleared the reduced diff on every count — `pdf-extract` back at 0.10.0, `lopdf` at 0.38.0, the three manifests identical to the base, `cargo metadata --locked` passing, and the earlier font-cache finding not surviving into what remains.

It also caught a false claim in my own justification, which is worth recording rather than quietly editing away.

I wrote that a local fix was impossible because `[patch.crates-io]` with a git source would block publishing `wenlan-core`. That is wrong. Cargo removes the entire patch table when it packages a crate.

Verified rather than taken on trust: I packaged a scratch crate carrying a `[patch.crates-io]` entry and read the manifest back out of the resulting `.crate` archive. It contains no `[patch]` section at all.

**The real limitation is different, and weaker.** A patch does not travel with the published crate, so anyone consuming `wenlan-core` from crates.io would still resolve the corrupting `pdf-extract` 0.12. And it would mean maintaining a fork of a PDF parser indefinitely.

That is a standing commitment rather than a one-off fix, so it is a decision to make deliberately, not a side effect of closing an advisory. The hold stands on the font-decoding regression itself, which is unchanged and remains the whole reason for it.

PR #626, opened 2026-08-29
docs: correct Windows signing guidance and document the SmartScreen warning

7xuanlu:
## Why

Two gaps, both about the Windows installer.

**The READMEs said nothing about SmartScreen.** All four tell macOS users what Gatekeeper will say and how to get past it. None mentioned that an unsigned `-setup.exe` makes Windows show "Windows protected your PC" with no visible way forward — the "Run anyway" button appears only after clicking "More info". A first-time user reasonably reads that as a broken download.

**`docs/code-signing.md` carried three claims that are no longer true.** It is the document we would act on when picking a Windows signing provider, so a wrong table there costs real money.

| Claim in the doc | What is actually true |
| --- | --- |
| An extended-validation certificate buys "immediate reputation" | Microsoft stopped granting them an automatic SmartScreen boost in March 2024. Ordinary and extended-validation certificates now earn reputation the same way, by download volume. The pricier tier buys stricter validation of who you are, not a quieter first install. |
| "Azure Trusted Signing … not offered in every country" | Renamed **Azure Artifact Signing**. Public-trust certificates go to organizations in twelve listed countries; an individual developer must be in the United States or Canada, with an Azure billing account of type "Individual". Identity validation takes 1 to 20 business days. |
| SignPath Foundation absent | It signs open-source projects for free and keeps the key in its own hardware module. The trade-off is that Windows names SignPath Foundation as the publisher, not Wenlan. |

## What changed

- All four READMEs gain one sentence in the Windows paragraph, in their own language, mirroring the existing macOS note: the installer is not signed yet, so choose "More info" and then "Run anyway".
- `docs/code-signing.md` section 2 is rewritten with the corrected provider table, and gains a paragraph saying plainly that no certificate removes the warning on day one.
- The summary table and intro no longer claim a certificate alone satisfies SmartScreen.
- Translation sync markers restamped (`scripts/check-readme-translations.py --write-markers`).

## What did not change

No provider is chosen and no signing is wired. `bundle.windows.signCommand`, `release.yml`, and `app/tauri.conf.json` are untouched. This PR only makes the provider choice an informed one and stops the unsigned installer from looking broken.

## Verification

```
$ python3 scripts/check-readme-translations.py
README translations are in sync
rc=0

$ bash scripts/check-readme-translations.test.sh
selftest ok
rc=0

$ git diff --check
whitespace rc=0
```

Sources for the corrected facts: [Microsoft Learn, Artifact Signing quickstart prerequisites](https://learn.microsoft.com/en-us/azure/artifact-signing/quickstart), [Trusted Signing opens to individual developers](https://techcommunity.microsoft.com/blog/microsoft-security-blog/trusted-signing-is-now-open-for-individual-developers-to-sign-up-in-public-previ/4273554), [SignPath Foundation](https://signpath.org/).


7xuanlu:
Adversarial review returned **DO NOT MERGE** on `58f7a5d7` with three P1s and four P2s. All seven are addressed in `fefc8012`.

**P1 — the Simplified Chinese button label was wrong.** The review flagged `“详细信息”` and suspected the dialog says `“更多信息”`. It does. The SmartScreen dialog reads "Windows 已保护你的电脑 / Microsoft Defender SmartScreen 阻止了无法识别的应用启动", and the link that reveals the run button is **更多信息**, then **仍要运行** ([Microsoft Q&A, zh-CN](https://learn.microsoft.com/zh-cn/answers/questions/4263946/windows-windows-microsoft-defender-smartscreen)). A wrong label is worse than no instructions, because the reader cannot find the button at all. Fixed.

**P1 — the SmartScreen and Azure claims were unsourced.** Both are now sourced to [Microsoft's *Code signing options for Windows app developers*](https://learn.microsoft.com/en-us/windows/apps/package-and-deploy/code-signing-options), linked inline, with a line saying that page wins wherever it disagrees with our table. Microsoft states the extended-validation bypass "was removed in 2024" and that paying the premium for SmartScreen alone is "no longer justified", so the over-precise "March 2024" is gone. The twelve-country list is replaced by a link to Microsoft's prerequisites; only the individual restriction (United States or Canada) stays inline, because that one decides whether this project can use the service at all. Prices now match Microsoft's figures: USD 150–300 for an ordinary certificate, USD 400+ for extended validation.

**P2 — the summary table overpromised.** The Fix cell said signing fixes a row whose symptom is the SmartScreen warning, contradicting the section below. It now says signing establishes a publisher identity and that the warning fades only as that identity earns reputation.

**P2 — the wiring recipe contradicted itself.** It asked the sign script to exit 0 when secrets are absent and then required `Get-AuthenticodeSignature` to be `Valid`, which would fail every unsigned fork build. The verification step is now guarded the way the macOS one is guarded by `APPLE_SIGNING_CONFIGURED`.

**P2 — the recipe left stale docs behind.** It never told the maintainer to update the summary table and the four READMEs after signing is enabled, so they would keep saying "not signed yet". Section 1 already required that cleanup for macOS; section 2 now does too.

**P2 — SignPath Foundation was overstated.** It no longer guarantees whose name the certificate carries, and points at their eligibility terms instead.

**P3 — quoting conventions.** Spanish used guillemets where that file uses straight quotes; Simplified Chinese used `“…”` where that file uses `「…」`. Both now match the macOS sentence two paragraphs above.

**One addition the review did not ask for.** Microsoft's page names a free, warning-free path this document had never considered: publishing an MSIX package through the Microsoft Store, which Microsoft re-signs. It is out of reach — Tauri bundles NSIS and MSI, not MSIX, and a Store submission of an MSI or EXE installer must still be signed by the publisher — but the table now records that, so nobody re-derives it later.

**Not accepted as written:** the review asked for the eligibility country list to be sourced rather than removed. Two Microsoft pages disagree on the organization list (the quickstart names twelve countries; the code-signing-options page names four), so reproducing either here would be the drift the finding warns about. Linking Microsoft's prerequisites is the safer form of the same fix.

Verification after the fixes:

```
$ python3 scripts/check-readme-translations.py
README translations are in sync
rc=0

$ git diff --check
whitespace rc=0
```

PR #627, opened 2026-08-29
fix(release): notarize the DMG so a first launch shows no dialog

7xuanlu:
The six Apple secrets are now set, so the next release is the first one signed with a Developer ID certificate. Two things stood between that and a new user opening the app without a Gatekeeper dialog.

## The disk image was never notarized

Tauri notarizes and staples the `.app`, but the `.dmg` it wraps around that app is only **signed** — see `sign::sign` with no notarize call in [`crates/tauri-bundler/src/bundle/macos/dmg/mod.rs`](https://github.com/tauri-apps/tauri/blob/tauri-cli-v2.11.4/crates/tauri-bundler/src/bundle/macos/dmg/mod.rs#L193-L207) at the CLI version this repo pins (`@tauri-apps/cli` `^2.11.4`).

That matters because the `.dmg` is what a browser download hands a new user, with `com.apple.quarantine` set on it. Gatekeeper assesses the disk image itself on the first double-click, and a signed-but-unnotarized image is not accepted there — exactly the friction the certificate was bought to remove.

`release.yml` now submits the `.dmg` to the notary service after the build, staples the ticket so the check works with no network, and asserts both the app and the image:

```
xcrun stapler validate "$dmg"
spctl --assess --type open --context context:primary-signature --verbose=2 "$dmg"
```

The macOS bundle job goes from 90 to 120 minutes, because notarization now waits on Apple's queue twice. That wait is real: a throwaway pre-flight submission today sat `In Progress` for over 30 minutes with Apple's system status page reporting the Notary Service healthy.

## The first-run gauntlet could not fail on this

`scripts/first-run/macos-app.sh` recorded the signature and the Gatekeeper verdict with `info`, which never fails a run. That is why ad-hoc signing shipped release after release and only surfaced as a hand-written finding.

Two things were missing, and both are fixed:

- **No quarantine attribute.** `curl` sets none, and that attribute is the only reason Gatekeeper is consulted at all, so the old assessment ran on a quieter path than a real user's. The script now stamps it by hand before assessing.
- **No assertion.** The checks are now hard, and the one that matters is `source=Notarized Developer ID`. That is the only verdict meaning a first launch opens with no dialog — signed-but-unnotarized is rejected, and ad-hoc signed is rejected with no source line at all.

Note that running the gauntlet against a tag from before this change will now fail these checks. That is the intended reading: those releases really do stop a new user with a dialog.

## Verification

```
$ actionlint .github/workflows/release.yml
rc=0

$ bash -n scripts/first-run/macos-app.sh
rc=0

$ shellcheck scripts/first-run/macos-app.sh
SC1091 (info): Not following: lib.sh was not specified as input   # pre-existing
rc=0
```

The end-to-end proof is the next release plus a gauntlet run on that tag, which is the step after this merges.


7xuanlu:
Adversarial review returned **DO NOT MERGE** on `9359af9d` with one P1, two P2s, and one P3. All four are addressed in `17947fa3`.

**P1 — the change certified the wrong outcome.** The comment said `source=Notarized Developer ID` is "the only source string that means a first launch opens with no dialog at all". That is wrong, and I confirmed it against Apple: "The first time that you open a new app from an identified developer that you downloaded outside the App Store, your Mac asks if you're sure that you want to open it" ([support.apple.com/102445](https://support.apple.com/en-us/102445)).

What notarization actually removes is the "Apple cannot check *App* for malicious software" refusal, whose only buttons are **Move to Trash** and **Done**, and the detour through System Settings to approve the app. The one-time "downloaded from the Internet, are you sure?" confirmation stays, and it has an **Open** button. The comment now says that, and adds that `spctl` assesses policy rather than replaying a launch.

I did not add a real download-drag-launch test. That needs a window server session and a human-visible dialog on a headless runner, and it would be flaky in exchange for confirming a confirmation box that Apple documents. The honest framing plus the policy assessment is the proportionate fix. The one path that genuinely shows nothing is the README one-liner, which strips the quarantine attribute, and that is now stated in the docs.

**P2 — the stand-in quarantine value was not browser-shaped.** It was `0083;00000000;Safari;`, with a zero timestamp and no UUID. It is now built as `0083;<hex epoch>;Safari;<uuidgen>`, the four-field shape Safari writes. The comment also now says plainly that writing the attribute marks the file without recording the LaunchServices quarantine event a browser would, so it stands in for a download rather than reproducing one.

**P2 — the owning doc became false on merge.** `docs/code-signing.md` said the DMG "is Developer ID signed but never notarized or stapled itself" and to notarize it only "if users report macOS blocking the DMG on open" — exactly what this branch changes. That bullet is rewritten, steps 6 and 7 name the new workflow step and the new check names, and the stale "notarization adds 2–15 minutes" line is replaced with the wait actually measured today.

**P3 — the tag cutoff was unstated.** The gauntlet's `release_tag` input now says the macOS leg asserts signing and notarization, which no release before v0.17.4 has, so an older tag fails that leg by design.

**Cleared by the review, worth recording:** the Tauri claim holds at the pinned CLI 2.11.4; `spctl -a -t open --context context:primary-signature` is Apple's documented DMG assessment and `accepted` is the right substring; stapling happens before the checksum step at `release.yml:665`, so published hashes cover the stapled DMG; `APPLE_ID` written to `$GITHUB_ENV` is visible to the later step's `if` condition, and a fork with no secrets still builds.

Verification after the fixes:

```
$ bash -n scripts/first-run/macos-app.sh
rc=0

$ shellcheck -x scripts/first-run/macos-app.sh
rc=0

$ actionlint .github/workflows/first-run-gauntlet.yml .github/workflows/release.yml
rc=0
```

Still unproven here, by design: live notarization of a real DMG and the Gatekeeper verdict on it. That comes from the first signed release plus a gauntlet run against its tag.

PR #628, opened 2026-08-29
ci: read Apple's notarization log on every verdict

7xuanlu:
`xcrun notarytool submit --wait` prints `status: Accepted` or `status: Invalid` and nothing about why. Apple keeps the reasons in a per-submission log that has to be fetched separately — this is what Apple's *Resolving common notarization issues* tells you to read. Today a rejected release build fails with no usable detail, and finding out why costs another hour-long run.

Two additions, both diagnostic only:

- The **Notarize and staple the DMG** step now fetches `notarytool log` for its own submission on every verdict, accepted or not. Accepted submissions can still carry warnings worth seeing.
- Tauri submits the `.app` to the notary service from inside `pnpm tauri build`, where there is no hook. A new step reads the queue back with `notarytool history` and prints the log for the three most recent entries, but only when the build failed and the Apple secrets are present.

Neither step can turn a red build green: the history dump runs under `if: failure()`, and the DMG log fetch is followed by the same exit-code and `status: Accepted` assertions as before.

While writing this I audited the whole macOS pipeline against Apple's published guidance. It conforms on every point I could check against the pinned bundler source at `tauri-cli-v2.11.4`: sidecars and frameworks are signed individually before the app bundle, inside out, with `--options runtime` and without `--deep`; the disk image is signed, notarized and stapled as the outermost container; and the updater tarball is built after the app is stapled, so the archived app carries its own ticket as Apple requires for archives that cannot be stapled themselves.

`actionlint` is clean and `scripts/release-workflow-contract.test.py` passes.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -588,6 +588,24 @@ jobs:
           TAURI_SIGNING_PRIVATE_KEY_PASSWORD: ${{ secrets.TAURI_SIGNING_PRIVATE_KEY_PASSWORD }}
         run: pnpm tauri build --target aarch64-apple-darwin
 
+      # Tauri submits the .app to the notary service from inside `tauri build`
+      # and reports only that notarization failed, never Apple's reasons. That
+      # submission is invisible from here otherwise, so on a failed build read
+      # the queue back and print the log Apple keeps for each entry.
+      - name: Read Apple's notarization verdicts after a failed build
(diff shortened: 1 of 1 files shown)


7xuanlu:
Its commit rode inside #629, which merged as 6adf9fd6. `main` now reads Apple's notarization log on every verdict, so there is nothing left here to merge.

PR #629, opened 2026-08-29
ci: notarize only the outermost container

7xuanlu:
Stacked on #628 — base it on `main` once that merges.

Apple's instruction for a nested layout: *"if you have an app inside an installer package on a disk image, sign the app, sign the installer package, and sign the disk image, but only notarize the disk image."*

We were submitting twice. `tauri-bundler` submits the `.app` from inside `pnpm tauri build` as soon as it can see `APPLE_ID`, `APPLE_PASSWORD` and `APPLE_TEAM_ID` in the environment, and `release.yml` then submitted the `.dmg` separately. This withholds those three from the build environment — the bundler logs `skipping app notarization` and carries on with a signed app — so the disk image becomes the only submission. The steps that do talk to the notary service now take the secrets from `secrets` directly instead of the job environment, which is tighter scoping anyway.

**The updater archive is rebuilt.** A `.tar.gz` cannot hold a stapled ticket, so Apple's rule for archives is to staple each item you put inside one. Tauri writes `Wenlan_aarch64.app.tar.gz` before any ticket exists, so a new step staples the app (a lookup against the disk image's ticket, not a second submission), repacks the archive, and re-signs it with `tauri signer sign` — which calls the same `sign_file` helper the bundler uses, so the updater sees the signature format it expects. The step then unpacks its own output and fails the job unless the app inside passes `codesign --verify --deep --strict` and `stapler validate`. A silently broken updater archive is the worst thing this job could ship, so it fails loudly instead.

**One consequence worth knowing.** The app a person drags out of the disk image carries no ticket of its own. The single submission covers it — a ticket is a list of code directory hashes, and that app is in the list — but the copy inside the image was made before the ticket existed. Gatekeeper resolves it against Apple on first launch, which needs a moment of network. Everything a person keeps is stapled: the disk image they downloaded, and the app the updater installs. The verify step now also mounts the image and asserts the app inside assesses as `source=Notarized Developer ID`.

Net effect: one wait on Apple's queue instead of two.

`actionlint` with shellcheck is clean and `scripts/release-workflow-contract.test.py` passes. `docs/code-signing.md` is updated: the ordered step list, the single-wait note, and three replacement bullets covering the outermost-container rule, the unstapled in-image copy, and the rebuilt updater archive.


7xuanlu:
Codex Sol reviewed this at `58ada077` and returned DO NOT MERGE with two findings. Both are addressed in `64978d85`.

## The real one: the notarization wait gave up at 30 minutes

Sol is right. The job carries `timeout-minutes: 120` and `docs/code-signing.md` said the job allows 120 minutes, but the step passed `--wait --timeout 30m`. `notarytool` exits at its own timeout while Apple keeps processing, so a submission that settles at minute 45 fails the job and the staple-and-repack chain never runs.

Raised to `--timeout 90m`. Not 120, because the build ahead of this step has already spent part of the job budget and a wait that outlives the job is killed before it can print anything useful.

Sol's sharper point was the second-order one: a rerun submits the same disk image again and queues behind the first. That is exactly how this account accumulated several unsettled submissions. So a non-zero exit now prints the submission id and says to wait on that id:

```
ERROR: notarytool exited 1 for the DMG.
Submission <id> is still with Apple. Wait on that id rather than resubmitting.
```

`docs/code-signing.md` now states the real numbers and that recovery path.

## The other one: not a bug, fixed anyway

Sol flagged that the disk-image verification does `rm -rf "$mount"` and then `hdiutil attach -mountpoint "$mount"` without recreating the directory, predicting `hdiutil: attach failed - No such file or directory`.

That is not what happens. Checked empirically on macOS 26.5.1 with a scratch disk image: with the mountpoint confirmed absent, `hdiutil attach -nobrowse -readonly -mountpoint` returned 0, mounted at that path, and the volume contents listed.

```
mountpoint exists before attach? no
attach rc=0
/dev/disk5s1  41504653-...  /private/var/folders/.../mnt
f.txt
```

But that check ran on macOS 26.5.1 and the runner is `macos-14`, so the version question is open. Adding `mkdir -p "$mount"` is one line and settles it either way, so it is in.

## What Sol confirmed

Worth recording, because these were the parts I could not prove from a live run:

- Withholding `APPLE_ID`, `APPLE_PASSWORD` and `APPLE_TEAM_ID` has no signing side effect. Tauri's certificate signing reads only `APPLE_CERTIFICATE`, its password, and the signing identity; the three withheld variables are read by `notarize_auth()` alone. This job defines no API-key credentials either, which is the other path Tauri could have taken.
- The rebuilt updater archive is shape-compatible with what the updater expects: one `<name>.app/` root, symlinks and modes preserved, and updater 2.10.1 discards that first path component while extracting.
- The re-signing step targets the rebuilt archive and writes the `.tar.gz.sig` that the later staging and `latest.json` steps consume.
- No gating failure is swallowed; the only `|| true` uses are the diagnostic log read and cleanup detaches.
- Nothing secret reaches the log, and the log read is scoped to this submission rather than team history.
- The unsigned fork path stays guarded: both new mutation steps require `APPLE_NOTARIZATION_CONFIGURED == 'true'`.

## Verification

- `actionlint .github/workflows/release.yml` — exit 0.
- `python3 scripts/release-workflow-contract.test.py` — PASS.
- The staple, repack and mount chain is still code that has never executed end to end, because Apple has not settled a submission for this account. Apple Developer Support case 20000151317315 is open on that.

7xuanlu:
## Confirmation review passed, plus one follow-up fix

Sol re-reviewed the two fixes at `64978d85` and returned **MERGE**.

**The mountpoint fix is in and correct.** `mkdir -p "$mount"` runs immediately before `hdiutil attach`. Sol accepted the empirical result — on macOS 26.5.1 `hdiutil attach -mountpoint` creates a missing mountpoint and exits 0 — and noted the fix makes the question moot either way, since the runner is `macos-14` and was never tested.

**The timeout fix is in and correct.** `rc=$?` sits directly after the `notarytool` assignment, so it captures that command's status rather than the assignment's, and `set +e` / `set -e` bracket only that one command. No credential reaches the log: neither the captured output nor `notarytool log` contains the password, and there is no `set -x` in the step.

Sol flagged one thing it could not test and said so plainly: it reasoned from `notarytool`'s output structure, not from a live run, that the `id:` line is printed during submission and therefore already captured before a `--timeout` exit. The empty case is guarded by `if [[ -n "$sub_id" ]]` regardless.

### The follow-up: `bfa9b94c`

Sol's one piece of pushback was a margin trade rather than a logic error, and it is worth closing now.

`app-bundle` compiles the whole workspace from scratch before it waits — no Cargo cache on that runner, only pnpm's — and staples, repacks the updater archive, verifies and stages after. With the job capped at 120 minutes and the notary wait raised to 90, a slow compile leaves the notarize step less than its own budget, and GitHub kills the job at the cap with no message. That is the original bug wearing a different hat: no submission id printed, and a live submission left in Apple's queue.

The job cap moves to 210, which covers a from-scratch compile, the full 90-minute wait, and the steps after it. `scripts/release-workflow-contract.test.py` pins that bound, so its expected value moves too, and the comment on both says the numbers have to be raised together.

Verified: `actionlint .github/workflows/release.yml` exits 0, and `python3 scripts/release-workflow-contract.test.py` passes. The contract test failed with `release job 'app-bundle' does not keep its 120-minute bound` before its expectation was updated, so the pin is real rather than decorative.

PR #630, opened 2026-08-29
docs: choose SignPath Foundation and publish the code signing policy

7xuanlu:
Windows installers ship unsigned today. SmartScreen shows "Windows protected your PC" with an unknown publisher, and because an unsigned file cannot carry reputation from one release to the next, that warning never fades on its own.

This picks the provider and puts in place the one thing that has to exist on the project's home page before we can apply.

## Why SignPath Foundation

It is the only free option that puts a real Authenticode signature on the installer. Azure Artifact Signing costs about USD 10 a month and, for an individual developer, is limited to the United States and Canada. A commercial certificate is USD 150 or more a year. Sigstore does nothing for SmartScreen. Doing nothing leaves the warning permanently, because Microsoft is explicit that "reputation cannot transfer from previous versions unless both were signed using the same publisher identity".

Wenlan meets every published criterion at <https://signpath.org/terms.html>: an OSI-approved license with no commercial dual-licensing, no proprietary component, actively maintained, already released in the form to be signed, documented, an uninstaller, and a build on a trusted build system — GitHub-hosted runners qualify, and `app-bundle-windows` already runs on `windows-2022`.

## Two consequences worth knowing before we apply

**The certificate is issued to SignPath Foundation, not to Wenlan.** Their terms: "The code signing certificate is issued to *SignPath Foundation*. This means that *SignPath Foundation* is the publisher of the OSS project." Windows will name them as the verified publisher. The upside is that the same certificate already signs several hundred projects, so it arrives with reputation a fresh certificate would not have.

**A maintainer approves every release.** Their terms require that "each signing request must be approved by a team member". The Windows job will block on that approval, so a tag release stops being fully unattended.

## What this changes

- A `Code signing policy` section in all four READMEs, with the exact heading and credit line their terms require, naming the Author, Reviewer and Approver roles and stating the multi-factor authentication and privacy positions. The Foundation checks the home page for this during review, so it needs to be on `main` before the application goes in.
- Section 2 of `docs/code-signing.md` rewritten around the decision: what to do before applying, what the release workflow needs afterwards, and why the alternatives lost.
- One correction: the docs pointed at `app/tauri.conf.json` for Windows bundle settings. The `bundle.windows` block actually lives in `app/tauri.windows.conf.json`, the Windows overlay.

## What is deliberately not here

No workflow changes. Wiring the signing step needs an organization id, a project slug, a signing-policy slug and an API token, and none of those exist until the Foundation accepts the application. The plan for that wiring is written down in the doc, including the trap that matters most: the updater `.sig` beside the installer is computed over the unsigned bytes, so signing invalidates it and it has to be regenerated — the same repair the macOS path already makes after stapling.

The application text for the human to submit is drafted outside the repository.

## Verification

- `python3 scripts/check-readme-translations.py` — in sync, after restamping the markers with `--write-markers`.
- `bash scripts/check-readme-translations.test.sh` — `selftest ok`.
- `git diff --check` — clean.
- Every quoted requirement was read from <https://signpath.org/terms.html> directly, not from memory.


7xuanlu:
Superseded by #632, which is cut from `main` after #629 merged and contains only the six documentation files. This branch was stacked on the Apple notarization work, so it carried `.github/workflows/release.yml` and a throwaway `verify-apple-signing.yml` that must never reach main — the first blocker in the review of this PR. Every other finding from that review is addressed in #632.

PR #631, opened 2026-08-29
fix(app): stop cloudflared from replacing its own signed binary

7xuanlu:
## What

Pass `--no-autoupdate` when the app starts the cloudflared tunnel for Remote Access.

## Why

`cloudflared tunnel` leaves auto-update **on** by default. From the shipped CLI's own help text:

```
--no-autoupdate  Disable periodic check for updates, restarting the server with the new version. (default: false)
--autoupdate-freq value  Autoupdate frequency. Default is 24h0m0s.
```

So a running tunnel checks for a new version every 24 hours, downloads it, **overwrites its own binary**, and restarts.

Inside an installed Wenlan, that binary is not a loose tool. It is a bundled sidecar declared in `app/tauri.conf.json` and covered by the app's code signature. A self-update replaces a signed file with an unsigned one, on the user's disk, after installation, with nothing to tell them.

That defeats the point of signing the bundle at all:

- On macOS it breaks the app's signature and its notarization ticket no longer describes what is on disk.
- On Windows it means the installer's Authenticode signature vouches for bytes that have since been swapped out — while we are in the middle of applying for a code signing certificate precisely so that signature means something.

It also adds a periodic request to a Cloudflare update host that no Wenlan document discloses. `PRIVACY.md` is being rewritten in #630 to list every network request; this one should not exist rather than be documented.

Wenlan ships cloudflared on its own release cadence. Upgrading it is a Wenlan release, not something a sidecar decides for itself.

## The test

`is_expected_remote_tunnel_command` is what orphan cleanup uses to recognize the tunnel process. If it stops matching the argv we spawn, a tunnel survives the app quitting — a leaked process holding an unauthenticated public URL to the user's memory.

The new case is the exact argv `start_remote_access` now produces.

**Mutation control.** Narrowing the matcher to require `--url` at argument index 2 makes the new case fail:

```
thread 'remote_access::tests::orphan_cleanup_requires_the_exact_remote_tunnel_process_identity' panicked at app/src/remote_access.rs:1894:9:
assertion failed: is_expected_remote_tunnel_command("/tmp/cloudflared-aarch64-apple-darwin tunnel --no-autoupdate --url http://localhost:22000",
    22000)
```

The matcher was then restored and the suite re-run.

## Verification

- `cargo test -p wenlan-app --lib remote_access` — 29 passed
- `cargo clippy -p wenlan-app --all-targets` — clean
- `cloudflared tunnel --help` on the shipped CLI (version 2026.5.0) for the default quoted above

Not verified: the flag's effect on a live tunnel. Remote Access needs a real Cloudflare tunnel to exercise, and the change is a documented CLI flag rather than logic, so this is a reading of the interface rather than an observed run.

### DIFF
--- a/app/src/remote_access.rs
+++ b/app/src/remote_access.rs
@@ -586,6 +586,27 @@ fn wenlan_mcp_process_identity(pid: u32, port: u16) -> Option<String> {
     Some(format!("{}\n{}", started.trim(), command.trim()))
 }
 
+/// The exact argument vector `start_remote_access` hands to the cloudflared
+/// sidecar.
+///
+/// This lives in one place, rather than inline at the spawn site, so the
+/// regression test can drive the argv the app really spawns. When the two were
+/// written out independently, deleting `--no-autoupdate` from production left
+/// the test green -- the fix could be removed without anything failing.
+///
+/// Order matters as much as membership. cloudflared accepts `--no-autoupdate`
+/// before `tunnel` as well, but `is_expected_remote_tunnel_command` requires
+/// `tunnel` at index 1, so that placement would break the app's own recognition
+/// of the process it started and leave the tunnel running after a quit.
+fn remote_tunnel_args(port: u16) -> [String; 4] {
+    [
+        "tunnel".to_string(),
+        "--no-autoupdate".to_string(),
+        "--url".to_string(),
+        format!("http://localhost:{port}"),
+    ]
+}
+
 fn is_expected_remote_tunnel_command(command: &str, port: u16) -> bool {
     let args: Vec<_> = command.split_whitespace().collect();
     let Some(executable) = args.first() else {
@@ -1560,14 +1581,7 @@ async fn start_tunnel(
     // periodic request to a Cloudflare update host. Wenlan ships cloudflared
     // on its own release cadence; upgrading it is a Wenlan release, not
     // something the sidecar decides for itself.
-    let (mut tunnel_rx, tunnel_child) = match tunnel_command
-        .args([
-            "tunnel",
-            "--no-autoupdate",
-            "--url",
-            &format!("http://localhost:{}", port),
-        ])
-        .spawn()
+    let (mut tunnel_rx, tunnel_child) = match tunnel_command.args(remote_tunnel_args(port)).spawn()
     {
         Ok(child) => child,
         Err(error) => {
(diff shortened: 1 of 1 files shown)


7xuanlu:
## Review round 2 — the blocker is fixed at `01334876`

Adversarial review found the regression test proved nothing about the fix, and it was right. The spawn site built its argument list inline; the test wrote out a similar string by hand. Because the two were independent, **deleting `--no-autoupdate` from production left the test green**, and so did moving it ahead of `tunnel`. A silent revert of a signing-relevant change would have shipped.

### The fix

Both now come from a single `remote_tunnel_args(port)`. The spawn site calls it; the test asserts its exact contents and order, then feeds the result through the orphan-cleanup matcher.

Order is load-bearing, not just membership. cloudflared registers `--no-autoupdate` as a hidden root flag as well as a `tunnel` flag, so putting it before the subcommand still parses. But `is_expected_remote_tunnel_command` requires `tunnel` at index 1, so that placement would stop the app recognizing the process it started — leaving a tunnel running after a quit, holding an unauthenticated public URL to the user's memory.

### Mutation controls

Each was run against a `git diff --stat` showing the break actually landed in the file.

**Flag deleted from production:**

```
thread 'remote_access::tests::orphan_cleanup_requires_the_exact_remote_tunnel_process_identity' panicked at app/src/remote_access.rs:1914:9:
assertion `left == right` failed
  left: "--url"
 right: "--no-autoupdate"
test result: FAILED. 28 passed; 1 failed
```

**Flag moved before `tunnel`:**

```
thread 'remote_access::tests::orphan_cleanup_requires_the_exact_remote_tunnel_process_identity' panicked at app/src/remote_access.rs:1913:9:
assertion `left == right` failed
  left: "--no-autoupdate"
 right: "tunnel"
test result: FAILED. 28 passed; 1 failed
```

Restored: `cargo test -p wenlan-app --lib remote_access` 29 passed, `cargo clippy -p wenlan-app --all-targets` clean, `cargo fmt --check -p wenlan-app` rc=0.

### Confirmed by the same review, no change needed

- **Flag placement is correct.** cloudflared 2026.5.0 uses urfave/cli/v2; `--no-autoupdate` is installed on the `tunnel` command itself and its value reaches `NewAutoUpdater`, so `tunnel --no-autoupdate --url …` is valid and effective.
- **The matcher still matches** the new argument list.
- **No second spawner or matcher exists.** The only place cloudflared is executed is this function; the only command-line matcher is `is_expected_remote_tunnel_command`. Every other reference downloads, copies, declares, or documents the sidecar.
- **Users whose binary already self-updated are not made worse off.** The flag prevents future checks; it does not restore or re-sign bytes already replaced.

### Still unchecked

The effect on a live tunnel. The bundled sidecar in a checkout is a zero-byte CI placeholder, so no packaged artifact or real Cloudflare tunnel was exercised. This remains a reading of the command-line interface plus its upstream source, not an observed run.

PR #632, opened 2026-08-29
docs: choose SignPath Foundation and make the privacy policy true

7xuanlu:
Replaces #630, which was branched off the Apple notarization work and so carried `.github/workflows/` changes it had no business carrying. This branch is cut from `main` after #629 merged and touches six files, all documentation:

```
PRIVACY.md
README.md  README.es-ES.md  README.zh-Hans.md  README.zh-Hant.md
docs/code-signing.md
```

## What this does

**Picks SignPath Foundation** as the Windows code signing provider, and adds the one artifact that has to exist before applying: a `Code signing policy` section on the project's home page, naming the Author, Reviewer and Approver, with the exact credit line their terms require. That heading and that sentence are kept verbatim in English in all four READMEs, because SignPath looks for the literal term.

No workflow changes. The signing step needs an organization id, a project slug, a signing-policy slug and an API token that only exist once the Foundation accepts the application.

**Documents the two-license split.** The desktop app has shipped under AGPL-3.0-only since `app/LICENSE` was added, and both `package.json` and `app/Cargo.toml` declare it, but every README said "Wenlan is licensed under Apache-2.0" without qualification. Both licenses are OSI-approved and neither part is sold commercially, so the split satisfies SignPath's "OSI-approved license without commercial dual-licensing" — it just has to be stated, and the application answer has to name both.

**Rewrites the privacy policy so it is true.** This is the substantial part; see below.

## The privacy policy was wrong, and it mattered

The first version of this PR published SignPath's fixed sentence — that Wenlan "will not transfer any information to other networked systems unless specifically requested by the user". An adversarial review challenged it, and it does not hold. Three requests fire on their own, on an install that has configured nothing, and none of them had an opt-out:

| What | When | Where |
| --- | --- | --- |
| Search model download (~210 MB) | first daemon start | `huggingface.co`, `Qdrant/bge-base-en-v1.5-onnx-Q` |
| Desktop app update check | 3 s after every launch | `github.com/7xuanlu/wenlan/releases/latest/download/latest.json` |
| MCP server version check | every `wenlan-mcp` start, cached 24 h | `github.com/7xuanlu/wenlan/releases/latest` |

Each was read in the source, not recalled: `crates/wenlan-server/src/main/startup.rs:97` into `crates/wenlan-core/src/db.rs:4675`; `app/src/lib.rs:1280` into `app/src/updater.rs:107`; `crates/wenlan-mcp/src/main.rs:151` into `crates/wenlan-mcp/src/self_update_check.rs:87`.

`PRIVACY.md` also scoped itself to "the Wenlan daemon, CLI, MCP server, and Claude Code plugin" — the desktop app was not covered at all — and it blamed the Hugging Face download on the opt-in `wenlan models install`, which fetches a different model.

SignPath's terms permit **either** the fixed sentence **or** a link to a privacy policy. This takes the second option: the READMEs link `PRIVACY.md`, and `PRIVACY.md` now lists every network request with what it sends and how to avoid it, plus five things it never mentioned — the seven cloud providers the app offers presets for, the reranker's weight download, the test-endpoint and model-list buttons, Remote Access's 30-second health poll, and the persistent 16-byte identifier its relay registration sends as both account name and shared secret.

Telemetry stays "none". That one holds: there is no analytics library anywhere in the shipped code and the fonts are bundled rather than fetched.

## Six corrections to the Windows plan, all from the same review

1. **The workflow would have published the unsigned installer.** SignPath's action only downloads the signed file if `output-artifact-directory` is set — its manifest says "If not specified, the task will not download the signed artifact from SignPath" — and the existing staging step would then have picked up the original `*-setup.exe` from a green run. The plan now requires that input, requires overwriting the built installer in place, and requires the signature check to verify the publisher subject, not just the status.
2. **The updater re-signing step would have had no key.** `TAURI_SIGNING_PRIVATE_KEY` and its password are scoped to the build step alone.
3. **Bundled upstream binaries were told to stay out of the package.** That would have broken the app. SignPath's terms allow them: "You may include unsigned binaries of upstream OSS projects, e.g. DLL files, in your signed packages."
4. **Inner-binary signing was called speculative.** Microsoft documents that Smart App Control's "signature checks apply to all executable files, not just those downloaded from the Internet", so signing only the installer is a known gap to test, not a maybe.
5. **The certificate's reputation was overstated.** "Several hundred projects" appears in no SignPath source, and Microsoft says a newly created binary can warn even when signed.
6. **Two smaller overstatements.** The Foundation tier's consequences are not "irreversible", and the "few days to a few weeks" turnaround had no source.

## Verification

- `python3 scripts/check-readme-translations.py` — "README translations are in sync"
- `cargo test -p wenlan-core --lib drift_guard` — 156 passed, which covers the in-repo path references this PR moves around
- `python3 scripts/release-workflow-contract.test.py` — passes
- `git diff --name-only main...HEAD` — the six files listed above, nothing else

Every SignPath and Microsoft claim quoted here was read from the live page during this work.


7xuanlu:
## Review round 2 — blocker and four smaller findings fixed at `1c7d6f2b`

Adversarial review found a **fourth automatic network request** the rewritten policy still missed, which made the new text false in the same way the old text was. I verified every step of it against the source before acting.

### The fourth request: remote images in Markdown

Wenlan renders Markdown, and Markdown images take any address. Opening or editing a note containing `x` fetches that picture from that host.

| Where | What I read |
| --- | --- |
| `app/tauri.conf.json:35` | `"csp": null` — no content security policy blocks image loads |
| `src/components/memory/ContentRenderer.tsx:270` | passes `components` to `react-markdown` that override only `a`; `img` stays at the library default |
| `src/components/memory/editor/writingPresentation.ts:687` | `safeRemoteImageSource` explicitly admits `http:` and `https:` |
| `writingPresentation.ts:139-145` | assigns that URL to `image.src` |

The consequence is that a uniquely generated image URL works as a read receipt against anyone who opens a page you sent them, disclosing their IP address and the time they opened it.

`PRIVACY.md` now describes this under its own heading. It states plainly that it cannot be switched off and that **no local alternative exists** — the editor renders nothing but `http:` and `https:` images, so any picture in a note was fetched from somewhere. The telemetry section's "opening a window contacts nothing" is qualified for the same reason.

**Whether to block remote images by default is a product decision, not a documentation one, and it is deliberately not in this pull request.** This PR is documentation-only. Changing the behaviour belongs in its own change.

### Four smaller findings

1. **The third parties' own policies are now named.** SignPath's conditions say: *"Remember to specify the privacy policies of other Open Source or third party components or services your application uses if your users are affected."* I fetched that page and quoted it verbatim. A new table links GitHub, Hugging Face, Cloudflare and Anthropic, and states that the relay is ours rather than a third party's. All four links were checked for HTTP 200.
2. **"Ingests nothing automatically" was false** once a folder is connected — `app/src/lib.rs:1240` installs watchers, `:1262` syncs at startup. The policy now says the folder choice is the consent and each individual file is not a separate prompt.
3. **"No certificate removes the warning on a first release"** overstated Microsoft, who say a signed binary *could* warn. It now reads that a certificate does not guarantee a first release avoids one, matching the sourced wording earlier in the same file.
4. **Two unverifiable assertions.** Multi-factor authentication is now stated as the project's standing requirement in all four READMEs — something the repository can hold itself to — rather than as an audited fact about private account settings. The claim that SignPath publishes no service level became the checkable claim that its application page states no turnaround time.

### Confirmed fixed by the same review, no change needed

Seven of the nine earlier findings were confirmed fixed with evidence, including that the pull request really is documentation-only. The reviewer noted the correct base for that check is `origin/main`, not local `main`, which is stale — `git diff --name-only origin/main...HEAD` lists exactly the six documentation files.

The two-license direction was also checked and is clean: no Apache-2.0 crate links the AGPL app; the AGPL app depends on Apache-2.0 `wenlan-types`, which is the permitted direction.

### Verification

- `python3 scripts/check-readme-translations.py` — in sync, after restamping the markers
- `python3 scripts/release-workflow-contract.test.py` — passed
- Still exactly six documentation files

PR #661, opened 2026-09-03
chore(windows-track): integration view -- review the stack #662..#668

7xuanlu:
> ## ⚠️ Review the stack, not this PR
>
> This diff is 42,657 added lines across 85 files, which is not reviewable. It
> has been split into seven stacked PRs, each based on the one before it, so
> each shows only its own change. **Review and merge #662 → #668 in order.**
> This PR stays open only as the integration view: it is where CI runs against
> the whole branch at once.

| | PR | what it is | added | product | tests |
|---|---|---|---:|---:|---:|
| 01 | #662 | SignPath watch + Authenticode contract | 5,344 | 1,151 | 4,193 |
| 02 | #663 | host-process probes | 8,868 | 4,759 | 4,109 |
| 03 | #664 | first-run ownership before deletion | 4,850 | 4,335 | 515 |
| 04 | #665 | reading tri-state in the app | 9,566 | 1,700 | 7,866 |
| 05 | #666 | quick-capture halo | 460 | 182 | 278 |
| 06 | #667 | negative controls | 13,568 | 68 | 13,500 |
| 07 | #668 | Vulkan doc | 1 | 1 | 0 |
| | | **total** | **42,657** | **12,196** | **30,461** |

**71% of this branch is tests.** The product-code surface is 12,196 lines, and
the largest single PR by raw size (#667, 13,568) is 99% test harness.

The stack's tip is byte-identical to `windows-track` — `git diff` between them
is empty — and every level was checked to route cleanly through drift_guard's
Windows paths-filter rule, so no intermediate PR in the stack is red on it. The
three commits that fixed CI on this PR were folded into the levels that
introduced the defects they fix, rather than left trailing at the end, which is
why the stack has 11 commits to this branch's 10 (the routing fix splits in two:
each level routes the Windows-sensitive script it adds).

---

## What the branch is

Ten commits, one defect found ten times:

> **a failed measurement that is indistinguishable from a negative measurement**

A probe that crashes and is reported as "nothing is running". A check that could
not run and prints a pass. A fixture that failed to build and is graded clean. A
health poll that times out and returns `false`, meaning "unhealthy" rather than
"unknown". The remedy is the same shape every time: **pass / fail / unchecked,
with a named reason**, where `unchecked` is never a pass.

The defect kept reappearing inside the fix for the previous one, including
inside the verification tooling itself — see #662, where CI caught the new
Authenticode contract making exactly this mistake about its own fixture.

## Known residuals, not fixed

Named here so they are not mistaken for things nobody looked at:

- `scripts/dev-runtime.sh:1286` — microsecond-wide mkdir-lock race.
- `GAUNTLET_HEALTH_TIMEOUT_SEC` is read at `windows-zip.ps1:182-199` while
  `Wait-Health` hardcodes `-TimeoutSec 5`, so the knob does nothing.
- `Wait-Health` is still boolean, not tri-state.
- Four NSIS surfaces (`teardown-app-gone`, `teardown-sidecar-gone`,
  `uninstall-removes-registry-key`, `Get-UninstallEntry`) are covered by static
  analysis only.
- `Measure-LedgerBaseline` attributes by count rather than per-row stamp.
- Four `"."` path fallbacks at `crates/wenlan-core/src/config.rs:150,170,228,322`.
- `SIDECAR_STOP_LIMIT = 3s` against a measured ~2.03 s refused-connect latency.
- `std::mem::forget(WorkerGuard)` at `app/src/lib.rs:546`.
- `for i in $(seq 1 120)` still at `scripts/smoke-cli.sh:214` and
  `scripts/smoke-mcp.sh:212`.

`wenlan-core` cannot be built on the Windows machine this was developed on —
`llama-cpp-sys-2` exits 101 without the "Desktop development with C++" workload
in VS 2022 Build Tools — so anything requiring that crate to compile is CI-only
and is marked unchecked locally rather than assumed to pass.

https://claude.ai/code/session_01NH3nEZRDbNUeEzCrzfPuk8


7xuanlu:
## Negative-control sweep: the verdict promised in the PR body

`scripts/negative-controls/run-all.sh`, full sweep, on this branch:

```
registered=10 ran=10 ok=10 controls-failed=0 did-not-complete=0 skipped=0
  hasher: sha256sum reproduces a known SHA-256 over 53 fixture bytes
    (c2be5347acfeaeb0...), so it is hashing and not printing a constant
SUITE VERDICT: every registered harness completed and every control fired.
NEGATIVE-CONTROL COMPLETE run-all.sh failures=0 elapsed=5013s partial=0
```

Ten harnesses, ten ran, ten passed, every control fired, nothing skipped and
nothing left partial. 84 minutes.

**Scope, stated rather than implied:** that sweep ran against `81e2342f`, before
the three commits that fixed CI (`a5ca04f5`, `ce83ee6e`, `aedab1fd`). Two of
those three touch nothing the sweep covers — the ci.yml routing entry and the
two clippy edits. The third rewrites the Authenticode fixture logic, and
`authenticode-step-receipt.py` is being re-run against it now; until that
returns, that one harness is **unchecked on the current head**, not passed.

## The three CI failures on this PR, and what they were

All three were mine, and one of them was this branch's own defect class showing
up inside the tooling written to catch it.

| job | cause | fix |
|---|---|---|
| `linux-nextest` / `macos-nextest` | `drift_guard::ci_routing_is_fail_closed_and_differential` — two new Windows-sensitive script paths not routed through ci.yml's `windows` filter | `a5ca04f5`, now split across #664 and #667 so each level routes the file it adds |
| `app-check` | `needless_return` in `lifecycle.rs`, `redundant_closure` in `mcp_config.rs` | `ce83ee6e`, folded into #665 |
| `windows release contract` | the Authenticode truth table's tampered-installer row asserted `HashMismatch` and the runner produced `NotSigned` | `aedab1fd`, folded into #662 |

The third is worth reading. The row tampers with a signed binary and requires
the shipped release step to refuse it with `HashMismatch`. It passed here and
failed on `windows-2022`, and the row could not tell *"the gate is broken"* from
*"this fixture cannot ask that question"* — so it reported a violation.

Three assumptions underneath it, each measured rather than reasoned about:

1. `data[4096:4100] = b"ZZZZ"` is a patch only if the file exceeds 4 KiB. Past
   the end, Python slice assignment **appends** — silently turning the
   tampered-in-the-middle fixture into a second copy of the byte-appended one.
   The probe sorts by size ascending, so it goes looking for small files.
2. The probe took the first *valid* signed binary of either kind. Measured on
   this host: **613 of 622** validly signed `.exe` files under System32 and
   `$PSHOME` are **catalog-signed** — no embedded certificate table, so a
   modified copy reports `NotSigned` and can never report `HashMismatch`. Which
   kind the row received was luck. This host drew an embedded one at rank 1; CI
   drew a catalog one.
3. A row that exits for the right reason without proving its own claim was
   graded a failure.

The fix takes the offset from the file's own PE geometry, verifies the patch
actually changed the bytes, makes the embedded signature a *preference* with a
fallback (requiring it would answer `NONE-FOUND`, which this suite treats as a
broken probe rather than an incapacity), and reports a mismatch row refused with
some other named status as `unchecked` — carrying that status — while dropping
it from `ran` so the mutation loop stops crediting it.

Verified by reproducing the CI condition here rather than arguing it:

```
ARM 1  unmodified: mismatch runs, 0 violations, 0 unchecked
ARM 2  probe forced to return a CATALOG-signed fixture (what CI drew):
       0 violations, mismatch absent from `ran`, one unchecked line naming
       the missing certificate table
ARM 3  control: with the status check deleted from the shipped step, the
       mismatch row still reports a violation
```

ARM 3 is the one that matters — it is what keeps the concession in ARM 2 a third
state rather than a hole.

https://claude.ai/code/session_01NH3nEZRDbNUeEzCrzfPuk8

7xuanlu:
Closing: the stack is merged into main in full. #662, #663, #664, #666, #684 (readings, superseding #665, which auto-merged into a side branch and never reached main by that route), #667 and #668 landed in that order, each after its own CI on main. Main's tree now matches this branch's head bd631ad3 in every file the stack touched. The windows-track branch stays as the integration record.

PR #665, opened 2026-09-03
fix(app): a reading that failed no longer renders as a measured negative

7xuanlu:
Every commit on this branch is one defect, found again and again:

> **a failed measurement that is indistinguishable from a negative measurement**

A probe that crashes and is reported as "nothing is running". A check that could
not run and prints a pass. A fixture that failed to build and is graded clean. A
health poll that times out and returns `false`, meaning "unhealthy" rather than
"unknown". The remedy is the same shape every time: **pass / fail / unchecked,
with a named reason**, where `unchecked` is never a pass.

## What

The largest behavioural change in the stack. A reading the desktop app could not
take -- the daemon unreachable, a config file unparseable, a path that does not
resolve -- used to render as a **measured negative**: "not configured", "no
sources", "0 memories". It now renders as unknown, with the reason.

## The measurement problem it fixes

A user seeing "not configured" cannot tell whether the app looked and found
nothing or failed to look. The two need different actions from them, so
collapsing them is not a cosmetic bug.

## What was measured

82% of this diff is tests: `app/src/mcp_config.rs` alone adds 2,909 test lines
against 478 of product code, `lifecycle.rs` 1,743 against 1, `daemon_start.rs`
1,151 against 50. `src/test/readings.ts` gives the frontend the same three-state
shape, and `preview/mocks/live-invoke*` lets the preview harness produce the
failed-reading state on demand.

`cargo test -p wenlan-app --lib` is 503 passed. The 5 failures are pre-existing
and Windows-only: `api::tests` panics at `app/src/presence.rs:608` with
`SecretUnavailable`, because owner-only secret protection has no implementation
on this platform. CI passes them.

The second commit clears the two clippy errors CI found (`needless_return`,
`redundant_closure`); neither changes behaviour.

## Review history since first pushed

- Forty-eight single-case Rust tests folded into table-driven tests (about 880
  lines removed, no case dropped); a named type alias for the table row keeps
  clippy's `type_complexity` gate green on macOS.

---

Part of the `windows-track` stack, split out of #661 because 44,375 added lines
is not reviewable in one diff. Each PR below is based on the one above it, so
this diff shows only its own change. Merge in order.

| | PR | added | product | tests |
|---|---|---:|---:|---:|
| 01 | [#662](https://github.com/7xuanlu/wenlan/pull/662) SignPath watch | 5,656 | 1,206 | 4,450 |
| 02 | [#663](https://github.com/7xuanlu/wenlan/pull/663) host-process probes | 9,222 | 4,879 | 4,343 |
| 03 | [#664](https://github.com/7xuanlu/wenlan/pull/664) first-run ownership | 5,374 | 4,562 | 812 |
| 04 | [#665](https://github.com/7xuanlu/wenlan/pull/665) reading tri-state | 9,567 | 1,621 | 7,946 |
| 05 | [#666](https://github.com/7xuanlu/wenlan/pull/666) quick-capture halo | 460 | 182 | 278 |
| 06 | [#667](https://github.com/7xuanlu/wenlan/pull/667) negative controls | 14,096 | 41 | 14,055 |
| 07 | [#668](https://github.com/7xuanlu/wenlan/pull/668) Vulkan doc | 1 | 1 | 0 |

The stack's tip is byte-identical to `windows-track` (`git diff` between them is
empty), and every level was checked to route cleanly through drift_guard's
Windows paths-filter rule, so no intermediate PR here is red on it.

Reviewed twice by Codex Sol (adversarial) and once by a Fable gatekeeper on
the integrated tree; every finding was fixed in place and has a negative
control at level 6. The two Windows channels are never executed before merge:
their evidence is static reading, the 109-case Windows probe harness and the
35-case `lib.test.ps1` suite.

https://claude.ai/code/session_01NH3nEZRDbNUeEzCrzfPuk8

### DIFF
--- a/.claude/skills/prove/SKILL.md
+++ b/.claude/skills/prove/SKILL.md
@@ -6,8 +6,12 @@ description: Per-surface verification loops for wenlan — daemon, cli, mcp, plu
 # /prove [surface] — wenlan verification loops
 
 Run every check through the evidence wrapper (records to `.claude/attest.jsonl`,
-which the weekly sweep audits): `~/.claude/bin/attest.sh <command>`.
-Sandboxed local runs need `TMPDIR=/tmp/claude` prefixed (macOS mktemp gotcha).
+which the weekly sweep audits): `bash scripts/attest.sh <command>`. That wrapper
+ships in the repo and runs on macOS, Linux and Windows (Git Bash); the older
+`~/.claude/bin/attest.sh` is a personal copy that does not exist on every
+machine, and a missing wrapper means the run is unrecorded, which the sweep
+reads as "the smoke never ran". Sandboxed local runs need `TMPDIR=/tmp/claude`
+prefixed (macOS mktemp gotcha).
 
 | Surface | Command | Proves |
 |---|---|---|
--- a/.claude/skills/prove/references/mutprove.md
+++ b/.claude/skills/prove/references/mutprove.md
@@ -17,7 +17,7 @@ Scripted, never conversational (the model round-trip is the cost; measured
 
 ```bash
 # mutations.json: [{"id","desc","file","old","new"}, ...]
-TMPDIR=/tmp/claude ~/.claude/bin/attest.sh \
+TMPDIR=/tmp/claude bash scripts/attest.sh \
   python3 ~/.claude/bin/mutprove.py mutations.json -- \
   cargo test -p wenlan-core --lib <module>::tests
 ```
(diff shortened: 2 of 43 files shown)


7xuanlu:
This PR was squash-merged into the side branch `win-stack/05-quick-capture`, not into main: its base was retargeted while a stale auto-merge flag was still set, and side branches have no protection rules. The side branch has been reset to its intended content, so nothing from this PR reached main by that route. The same change is now #684, based on #666, which merges first because this level's SetupWizard test depends on the drag-strip behaviour #666 adds.

PR #737, opened 2026-09-12
feat(activity): show background work on the toolbar Activity button

7xuanlu:
Second PR of the activity-indicator arc. The daemon route landed in #736; this is the surface that reads it.

## What a user sees

The Activity button in the window toolbar, on every page (Settings and a collapsed sidebar included), carries the state in its icon. There is no number and no separate dot:

| State | Icon |
|---|---|
| Up to date | plain clock |
| Steeping | indigo clock, hands sweeping |
| Waiting for a quiet moment | indigo clock, hands still |
| Blocked | amber clock, hands become "!" |

Clicking it opens a summary by asset. Memories, entities and pages each get one sentence in their own unit, along with the model that does the work and a line about what does or does not leave the device. When a model is missing, the summary opens on that and the button that fixes it. "See all activity" goes to the Activity view, whose Now section shows the same assets with expandable steps.

The import surface reports the two phases it can speak for, Ingest and Store, then hands the rest to the Activity button: "3 memories stored and searchable now."

## Wire contract

`ActivityState` is one enum (`up_to_date`, `organizing`, `waiting_for_idle`, `blocked`), not a state plus a flag. Waiting means the scheduler's latest check held background work and no import-priority window could run it either.

Every enum in the response reads a word from a newer daemon as `unknown` instead of failing the read. The UI only says what it can back up: an unknown asset or step is left out, and no blocked cause is guessed for it. An unknown lane gets no chip and no models entry, and hides the trust line, because it may be a cloud vendor.

## Deviations from the brief

**No Retry control under Memories.** No daemon route reruns only failed enrichment steps. Failed rows retry on their own until `ENRICHMENT_MAX_ATTEMPTS`, then become `abandoned` (`crates/wenlan-core/src/post_ingest.rs:107-119`). The Blocked state, its count and its sentence ship; the control is deferred.

**Explicit a11y assertions instead of axe.** This follows `spaces-accessibility.spec.ts`: the dialog's role and name, focus return on Escape, keyboard reachability, and no unlabelled graphics.

## Review build

`get_activity` joins the Review command contract, since the toolbar reads it on every screen. The mock runtime answers it with a quiet up-to-date response.

## Spec

`~/.wenlan/specs/wenlan/2026-09-11-activity-indicator.md`


7xuanlu:
## /verify against a live daemon

Drove `GET /api/activity` on an isolated daemon (`:17049`, scratch
`WENLAN_DATA_DIR`, `WENLAN_NO_AUTOSTART=1`). The user's daemon on `:7878` was
never touched and is still healthy.

This PR is frontend-only (42 files, all `src/` and `e2e/`, zero Rust), so the
CLI and MCP smokes have nothing in it to exercise. The surface that matters is
the chain from the daemon's JSON to the rendered line, and the mocks in this PR
are exactly what that chain cannot prove about itself.

| Link in the chain | How it was checked |
|---|---|
| Daemon serializes the shape | Live `curl` on `:17049` |
| Tauri command reshapes nothing | `search.rs:434` returns `wenlan_types::activity::ActivityResponse`, the daemon's own struct |
| TS type matches the wire | Live payload compared field for field against `ActivityResponse` in `lib/tauri.ts` |
| React renders it | 125 Playwright tests in real Chromium, 7 inspected visual baselines |

Every top-level key, every asset key, every step key, all six step names
(`store`, `summarize`, `link`, `detect`, `confirm`, `write`) and the lane values
match.

### It found a real defect

The e2e fixture's **step** objects carried `blocked`. A step carries `failed`;
only the asset above it carries `blocked`. Nothing reads step-level failure
counts today, so the suite passed on a fixture that was quietly a fiction, and
the first feature to read one would have been written against `undefined`.

Fixed, and the fixture is now annotated `ActivityResponse` so the next drift is
a compile error rather than a green test.

### And a gate I had missed

`tsconfig.json` includes `src` only. `e2e/` lives in `tsconfig.tools.json`,
which CI runs as its own step, and Playwright transpiles specs with esbuild
without typechecking them. So an unused constant in the new spec passed tsc,
vitest and the full Playwright run, and **would have turned CI red**. Removed;
both configs are now clean:

```
pnpm exec tsc --noEmit                      → clean
pnpm exec tsc -p tsconfig.tools.json --noEmit → clean
```

### Residual risk

No screenshot of the live Tauri window: the app launches and runs (with an
isolated bundle identifier, since the release app holds the production one),
but this terminal lacks macOS Screen Recording permission, so
`driver.sh shot` returns `SCStreamErrorDomain Code=-3801`. Rendered evidence is
therefore Chromium, not WKWebView. For a text line and a popover the gap is
small, but it is a gap.

7xuanlu:
## Live-app verification (f1f5b60f)

Drove the real Tauri dev app against an isolated daemon (port 17049, scratch data dir and pages path; prod :7878 untouched) with 8 memories in the blocked, no-model state. Found and fixed five issues that no test had caught:

| # | Issue in the running app | Fix |
|---|---|---|
| 1 | Popover clipped by the sidebar's `overflow: hidden` | Pinned to both anchor edges; grid items can shrink; footer wraps as whole items |
| 2 | "No background work yet" shown while 16 items were blocked | "Nothing has run yet" |
| 3 | "no everyday model is loaded" printed under both Memories and Entities | Rows say "8 waiting for a model"; the cause is named once per missing model |
| 4 | Now rail read "no model loaded built in, no model" | A route with no model prints its lane alone |
| 5 | Status line was a full-width divided strip below the account card | Sits directly above the card in the footer gutter, nav-row styling, no divider |

Gates on this head: `tsc --noEmit` and `tsc -p tsconfig.tools.json` clean; vitest `src/` 2368 passed; full Playwright 125 passed (7 sidebar baselines regenerated for the move), then the activity and visual specs re-run after the last popover tweak (7 passed).

Open question: the status shows "Blocked 16" for 8 memories, because it sums blocked units across Memories (8) and Entities (8). Screenshots are not committed.

7xuanlu:
## 5d346218: status moves into the toolbar Activity button; Organizing becomes Steeping

Review feedback: the sidebar status line was a second Activity entry, and it disappeared when the sidebar was collapsed or when Settings replaced it.

| Before | After |
|---|---|
| Status line above the account card, plus a separate toolbar Activity button | One toolbar Activity button that carries the state |
| Always showed a state word | Shows only "Activity" when up to date; a dot and count appear when steeping or blocked; the accessible name reads "Activity, <state>" |
| Popover opened above the sidebar line | Popover drops down from the button, right-aligned, and closes on Escape or an outside click |
| "Open Activity" | "See all activity" |
| Organizing / 整理中 | Steeping / 沉淀中 / 沉澱中 (the wire value stays `organizing`) |

Both sidebars are back to their main versions. Before the first read the button is the plain Activity button, and it stays the same element once data loads, so keyboard focus is kept.

**Verified**
- Live dev app on the isolated :17049 daemon: closed state with "Activity • 16", and the dropdown with "See all activity".
- `tsc --noEmit` and `tsc -p tsconfig.tools.json` are clean.
- vitest src: 2369 passed. A mutation check showed the new outside-click test fails without its listener.
- Playwright:
  - The full run gave 121 passed and 4 failed. All 4 asserted the old copy or the old click-navigates behavior (`i18n` zh-Hans and zh-Hant, `import-flow`, `import-handoff`). After fixing them, those specs pass: 4/4.
  - `activity-status.spec.ts` passes 6/6 at 1487, 1280, 768 and 375px. It checks one status in the header and none in a sidebar, popover bounds, Tab order and outside-click close.
  - The `spaces-navigation` visual baselines were regenerated locally on macOS.

Open question still standing: at `ActivityStatus.statusDetail`, the badge's "16" adds Memories (8) and Entities (8) for the same 8 memories.

7xuanlu:
## db244c05: state without numbers, every count in its own unit

Memories, entities and pages are not 1:1. One memory can name several entities or none. The old badge summed memory-unit counts ("16"), and the Entities row printed memories scanned as entities, so the popover said "Entities 8" while Home said "Entities 0".

| Surface | Before | After |
|---|---|---|
| Toolbar badge | `Activity • 16` | `Activity` plus a pulsing indigo ring (Steeping) or a solid amber dot (Blocked). No number. The state word stays in the accessible name and tooltip. |
| Popover lead | "Some steeping is waiting on you." + "No everyday model is loaded. Choose one in Settings, Intelligence." | "Steeping is paused: no everyday model is loaded." + **Turn on a model** (Home's wording), which opens Settings, Intelligence |
| Entities row | `8`, "8 waiting for a model" | `0` (entities found, matches Home), "8 memories not yet scanned for entities" |
| Memories row | "8 waiting for a model" | "8 not yet summarized. All are searchable now." |
| Now step counts | "0 of 8" | "0 of 8 memories" |

Found but unconfirmed entities now read as settled ("21 found, 9 confirmed in the Wiki"), not as running. Confirm is the user's call.

### Verification (at db244c05's tree)
- `tsc --noEmit` and `tsc -p tsconfig.tools.json --noEmit`: clean
- vitest `src/`: 2373 passed, 1 failed. The failure was `EntitiesView > will not restore on a count a failed reload left stale`, which is untouched by this branch and passes alone (28/28). It ran while Playwright was loading the machine.
- Playwright full suite: **126 passed**. New spec: `a missing model leads the popover, in units that match each row`.
- Live dev app on the isolated daemon :17049 (blocked, no everyday model): the badge shows an amber dot with no number. The popover opens on the cause and Turn on a model, and the Entities row shows 0, matching Home.
- Light and dark renders of both badge states and the popover were checked.

7xuanlu:
Fixes from running the app with a model loaded (c8e10ccf, 4ebec4e0).

**1. A model turned on in Settings is now used.** Before, loading an on-device model, saving an Anthropic key or saving a provider in Settings didn't assign the model to any job, so Activity said steeping was paused even with a model running. Now Settings assigns the new model to any job that has none, using the same preference as onboarding (`src/lib/routingPins.ts`). A job that already has a model keeps it. The blocked message now tells two cases apart:
- "no everyday model is chosen" when no model is assigned
- "the everyday model is not available" when the assigned model is still loading or its server is down

**2. The Steeping message says why work can wait.** It now reads "It works while your computer is quiet, so it can pause while you use it." This matches the scheduler, which holds heavy work while CPU or memory is busy, and on macOS until there has been no keyboard or mouse input for a minute. The message doesn't name a platform.

**3. The entity counts agree.** Before, the popover showed "0 of 9 memories scanned" beside 5 entities. The count required two step records, but only one of the three code paths that find entities writes both. A memory now counts as scanned when entity extraction finished or the memory is linked to an entity. The "Last activity" time also includes page compiles, so it no longer says "Nothing has run yet" next to pages that were written.

**Verification**
- Both tsc configs pass.
- Vitest: 980 passed. A new component test fails when the fill call is removed.
- e2e `activity-status.spec.ts`: 7/7.
- `wenlan-core` activity_counts: 5/5. `wenlan-server` activity tests: 25/25. Pre-commit clippy passed.
- Live on the isolated dev daemon (never :7878):
  - While the model was loading, the popover said the model is not available.
  - After clearing the model assignments, it said no model is chosen.
  - Running the fill helper through the app's Tauri commands assigned `on_device` to both jobs.
  - With the model loaded, it showed the Steeping message, "3 of 9 memories scanned" beside 5 entities, and "Last activity 38m ago".
- Not live-checked: clicking Load in Settings (no input automation is available). The component test covers that click.

7xuanlu:
Independent review of `db244c05..4ebec4e0` by Astra (gpt-6-astra, xhigh) through acpx, plus a live run against the isolated dev daemon. Fix pushed in 835171de.

**Findings**
- **F2, fixed in 835171de.** When Settings fills an unassigned job, it could pick an on-device model that was selected but not loaded: the routing pool lists a model once it is selected (`config_routes.rs:440`). Example: a load fails, then you save an Anthropic key, and everyday gets the dead on-device model instead of Anthropic. The Settings fill now counts on-device only once it is loaded. Onboarding's deferred download is unchanged, and Load still picks the model because it returns only after loading. There are two new tests, and both fail against the previous helper.
- **F1, accepted as a known limit.** The fill reads the pins and then writes them, so a pin changed in the milliseconds between the two can be overwritten. A full fix needs a daemon-side "fill if unset" operation, which is out of scope here.
- **F3 and F4, existed before this PR.** A refinery scan that finds zero entities still counts as unscanned; the refinery also picks that memory again. And an old extraction record still counts a memory as scanned after it is edited, because the count ignores the memory's version.
- Claims held in the source: the chosen vs. not-available copy, the Steeping message across platforms (the CPU and memory checks run everywhere; macOS also waits for input idle), and the last-activity time.

**Live (isolated daemon :17049)**
- "3 of 9 memories scanned" matches independent daemon reads: 9 memories, and 5 entity links across 3 memories. The old query would have shown 0 of 9.
- Clearing the pins makes Activity report "blocked" (`unconfigured`). Re-pinning makes it available again.
- `last_activity_at` equals the page's `last_compiled` to the second.
- Pins were restored exactly.
- Not observed live: the Settings Load click and the popover and footer wording. Clicking through the app bridge timed out, and the other click method needs macOS Accessibility permission, which was not granted. Those screens were checked earlier in this PR with a temporary force-open.

**Receipts for 835171de:** `tsc -b` clean; vitest for routingPins, intelligence and SetupWizard: 172 passed; pre-commit and pre-push passed.

PR #753, opened 2026-09-14
fix(onboarding): recover when setup stalls or the background service never starts

7xuanlu:
## What this PR does
Removes four first-run dead ends where setup could stall with no way out: the background service never coming up, a model download that stops moving, a chat import that fails with a raw IPC string, and Setting-up rows that print raw errors. Ledger items 1, 4, 5 and 6 from the 2026-09-14 onboarding churn review.

- **Service never starts.** The wizard's daemon row gains a "Start the service" button (calls `start_daemon_sidecar`, waits up to 30 s for `get_wire_state` to report the service reachable, then re-probes; Retry disabled while a start is in flight). The app now listens to `origin-fallback-mode` and shows a persistent banner with the same start action and a dismiss. When the boot gate errored instead of answering, the wizard leads with a connection-problem title above the welcome copy instead of reading like a fresh install. "Starting Wenlan" names what it is waiting for after 15 s (retry budget untouched).
- **Download stalls.** A model row whose byte count has not moved for 90 s flips to a stalled state with Retry; the stall detector is armed only against a known uncached catalog entry so a cached model never trips it. The load phase has a 15 min deadline and the loaded poll reads `isError`. Retry and resolve reset the progress stamp so a retried download is not judged by the previous attempt's clock.
- **Chat import errors.** ImportFlow uses the same localized heading and `formatImportErrorDetail` as ImportView. Three consecutive poll failures switch the strip to a lost state with Retry instead of "refining" forever; a new import resets that state.
- **Row errors.** New `describeSetupError` maps connection, timeout, permission and disk failures to localized copy and keeps the raw text as a detail line.

New i18n keys under `boot.*`, `setup.settingUp.*`, `setup.errors.*` and `chatImport.importFlow.*` in en, zh-Hans and zh-Hant.

Spec: `~/.wenlan/specs/wenlan/2026-09-14-churn-pr1-setup-dead-ends.md` (private working-doc home).

## How to test
Automated, at cea9d02c (rebased onto main after #751 and #752):
- `pnpm vitest run src/App.test.tsx src/components/SetupWizard.test.tsx src/components/ChatImport/__tests__ src/lib/bootRetryPolicy.test.ts src/lib/setupErrors.test.ts` → 7 files, 187 tests pass
- `pnpm test:i18n` → 57 pass
- `pnpm tsc --noEmit` and `pnpm exec tsc -p tsconfig.tools.json --noEmit` → clean

[...]

PR #754, opened 2026-09-15
fix(ingest): gate quick capture at ten characters like the memory store route

7xuanlu:
## Summary

- `POST /api/ingest/memory` (the path the quick-capture window uses) accepted any non-empty content, while `POST /api/memory/store` (CLI and MCP capture) refuses anything under ten characters. A two-character quick capture therefore created a memory that every other capture surface would have rejected. The ingest route now applies the same floor, counted in characters rather than bytes, so ten Chinese characters pass and nine ASCII characters do not.
- Quick capture checks the same floor before invoking and shows "Write at least 10 characters before saving." (zh-Hans, zh-Hant added) instead of round-tripping a server error. The alert clears as soon as the draft changes. The visible character counter now counts the same way the gate does.

Follow-up 2 from the onboarding churn ledger (after #753). No handler manifest change: the existing handler is edited in place.

## Verification (branch head 892e09f7)

- `cargo test -p wenlan-server content_length_gate_tests`: 4 new tests pass (nine ASCII characters rejected, ten Chinese accepted, four and nine Chinese rejected); the module's 8 tests run green.
- `pnpm vitest run src/components/QuickCapture.test.tsx`: 7 pass, including the new "blocks a too-short draft client-side without invoking".
- `pnpm test:i18n`: 57 pass. `pnpm tsc --noEmit`, `cargo fmt --check`, `cargo clippy -p wenlan-server`: clean.
- Rendered check in the worktree dev app (isolated daemon on :17530, app id `com.wenlan.desktop.dev.3511783530`): typed the nine-character draft "too short" into the real quick-capture window and submitted. The window shows the red alert "Write at least 10 characters before saving." under the draft, the draft is kept, the counter reads "9 chars", and no memory was created. Screenshot stored locally at `/tmp/claude/fu/qc-04-tooshort-en.jpg` (before: `qc-02-typed.jpg`); image upload from this session is unavailable, so the file is the evidence. The zh locales were checked by string only.

## Review

Independent Muse Spark review (read-only acpx session): verdict "ship with nits". Findings and closure:

- Nit: the visible counter used UTF-16 units while the gate counts code points. Fixed in 892e09f7.
- Nit: zh strings said 个字 / 個字 while the counter says 个字符 / 個字元. Fixed in 892e09f7.
- Nit: test comments implied ingest previously had a byte gate; rejection tests did not pin the message text. Fixed in 892e09f7.
- Noted, not changed: the ingest gate runs before the DB snapshot while the store route snapshots first, so a DB-down plus short-content request answers 422 here and 503 there. Harmless and cheaper.

Spec (private working doc): `~/.wenlan/specs/wenlan/2026-09-14-quick-capture-length-gate.md`.


7xuanlu:
Review closure. Muse Spark's independent read-only review returned "ship with nits": counter unit mismatch, zh unit wording, and test comment and assertion nits. All three were fixed in 892e09f7 and the checks listed in the description were rerun at that head. The rendered too-short state was driven in the real quick-capture window of the worktree dev app; the screenshot is kept locally because image upload is unavailable from this session. Ready for review; not merging without explicit approval.

7xuanlu:
**Closure re-check (Muse Spark via acpx, read-only, at 892e09f7)**

Verdict: closure ok. All three review findings confirmed resolved, no new defects.

1. `QuickCapture.tsx` counter now uses `Array.from(content.trim()).length`, matching the gate's units.
2. zh-Hans and zh-Hant copy read "10 个字符" / "10 個字元", matching each locale's counter.
3. `ingest_routes.rs` rejection tests assert the exact message text and the comments no longer claim a false history.

Awaiting merge approval.

PR #756, opened 2026-09-15
fix(app): fail the boot gate closed when the background service is down, even while the window is hidden

7xuanlu:
## Why

Seen live during the #753 verify: after the daemon died mid-session, "Starting the local runtime..." stayed pending for more than five minutes, while a plain reload with the daemon down errored within about 70 s as designed. Spec: `~/.wenlan/specs/wenlan/2026-09-14-boot-gate-stays-pending.md`.

## Root cause

React Query continues a retry only while `document.visibilityState` is not `"hidden"`, and an occluded or hidden WKWebView window reports hidden. A user who launches Wenlan and switches to another window parks the gate until they look at it again; it never fails closed on its own. Confirmed live with a temporary debug line on the boot screen: `visibilityState=hidden fetchStatus=paused failureCount=2` ten minutes after the reload, and one attempt every ~3 s as soon as the window was visible.

## What changed

- `runBootGate` in `src/lib/bootRetryPolicy.ts` owns the retry loop with a plain `setTimeout` delay and never consults document visibility or online state. The gate query sets `retry: false`. Budget arithmetic is unchanged (about 162 s worst case).
- Each attempt is bounded by `ATTEMPT_TIMEOUT_MS` on the frontend, and `should_show_wizard` is wrapped in a 6 s `tokio::time::timeout` on the Rust side, so a hung IPC call costs one attempt instead of the whole gate.
- Tests: `runBootGate` fast success, policy delays and last-error rethrow, and budget bound; App gate errors closed with a never-settling invoke and with the window hidden.

## Verification (at 643ce6e5)

- `pnpm tsc --noEmit` clean.
- vitest: App and bootRetryPolicy (36), SetupWizard and i18n guards (159) pass.
- `cargo fmt --check`, `cargo clippy -p wenlan-app`, `cargo test -p wenlan-app search` (69) pass at 33deec3e; the Rust side is unchanged since.
- Live, isolated dev app (port 17335, own data dir): daemon killed 45 s after launch, webview reloaded, window hidden for the whole wait. Before the fix the boot screen was still up after 240 s and after 10 minutes. After the fix, the wizard's "Could not reach the background service" state was already rendered when the window was focused at reload+100 s.

## Rendered evidence

Local screenshots only (`/tmp/claude/bootgate/40-prehide.jpg` before, `50-postfix-t100.jpg` after); image upload is not available from this session, so they are not attached. The after state is the existing wizard connection-problem screen from #753, unchanged by this PR.


7xuanlu:
Review closure. Muse Spark's independent read-only review returned "ship with nits" and checked, with no findings, the schedule equivalence (retry plus one attempts, same delays, same budget), the React Query pause surfaces, the Rust lock scope and error mapping, and the test mocks. Three nits, all applied in 3c037a2c:

- The retry sleep could not be cancelled, so an unmounted query kept firing IPC until the budget ran out. `runBootGate` now takes the query's AbortSignal and stops sleeping when it aborts.
- `should_show_wizard` used an inline state read where the `daemon_client` helper already existed.
- No test pinned the hidden window plus hanging attempt combination. Added one alongside an abort-mid-delay test.

At 3c037a2c: `pnpm tsc --noEmit` clean, `vitest run src/lib/bootRetryPolicy.test.ts src/App.test.tsx` 38 pass, `cargo check -p wenlan-app` clean. Ready for review; not merging without explicit approval.

7xuanlu:
**Closure re-check (Muse Spark via acpx, read-only, at 3c037a2c)**

Verdict: closure ok. All three review nits confirmed resolved, no new defects.

1. `app/src/search.rs` `should_show_wizard` now uses `daemon_client(&state).await`.
2. `bootRetryPolicy.ts` and `App.tsx` thread the query signal into an abortable delay that clears its timer and throws before the next attempt.
3. `App.test.tsx` pins the hidden `visibilityState` plus hung-first-attempt case and fails closed.

Reviewer note on abort semantics: an abort during the delay rejects with the last attempt error, but the retryer has already settled via cancel so the late rejection is ignored. An abort during an in-flight attempt takes effect at the next delay boundary with no extra attempt.

Awaiting merge approval.

7xuanlu:
Live proof at 3c037a2c, recorded with `bash scripts/attest.sh` (ledger status 0): isolated dev app on port 17335, daemon killed with ownership confirmed, webview reloaded, window hidden for 100 s. On focus the wizard already showed "Could not reach the background service" and stayed there at +110 s. Merged as 413e18a2.

PR #766, opened 2026-09-17
feat(sources): import OKF bundles as a source type that keeps provenance and links

7xuanlu:
## What this adds

An OKF bundle, the folder layout OpenWiki writes, can now be registered as its own source type instead of a plain folder. Each concept file becomes one source page, the whole bundle lands in one Space, the bundle's own frontmatter is kept beside the page for display, and the markdown links between concepts become real page links that resolve as their targets arrive.

Adding such a folder as a `directory` source today loses all of that: provenance is dropped, pages scatter across Spaces by tag, and the links between concepts never become links.

Third PR of the OKF arc, after #762 (export) and #763 (export hardening).

## How it works

**Parser** (`crates/wenlan-core/src/sources/okf.rs`, new). Reserved names (`index.md`, `log.md`, `INSTRUCTIONS.md`) are never concepts. Frontmatter becomes an `OkfConcept` with the block kept as parsed, so a shape Wenlan does not model is still stored rather than dropped. Links are `[text](target.md)` outside fenced blocks and inline code, normalized to bundle-relative concept ids. Documents come out in the same shape the folder parser produces, so `source_agent` stays `"folder"` and the existing folder immutability and citation rules apply unchanged.

**Storage** (migration 131). `okf_concepts` and `okf_concept_links` hold the provenance and the link set. The merge happens inside `refresh_page_wikilinks`, so every write path (worker, lint repair, rename, drafts) yields the same links and no path can silently drop them. A link resolves within its own source and Space, exactly or by a unique case-fold, and retires when its target goes away.

**Worker**. A source profile resolved from the config drives the OKF parse, the Space, and the concept's own `description` as the summary, with or without an LLM provider, so the same bundle does not summarize differently depending on the provider. An existing page's Space wins over the source's, so a page the user moved stays moved and its chunks follow it.

**Read surface**. `GET /api/pages/{id}` carries an optional `okf` object beside the page: the source id, the concept id, the frontmatter as parsed, and when it last changed. The key is absent for every other page, so a reader that does not know about OKF sees today's shape. No new Tauri command.

**Sync**. Reserved names and Wenlan's own export subtrees are pruned before anything is counted. An export root or a deleted Space marks the source unavailable rather than syncing. A deprecated concept goes through the existing removal path, which marks citing pages `source_removed`. A same-bytes rename rekeys the concept instead of re-importing it. At most 1,000 concepts are handed to the worker per sync, and the rest are reported as waiting rather than dropped. A file the worker eventually gives up on is counted as a sync error under `document_enrichment_failed`, so a bundle that imported nothing cannot report `errors: 0`.

**Registration**. A folder is refused when it is not a directory, when it is a bundle Wenlan exported, when it has no OKF signal, or when it overlaps the pages folder or another source in either direction. Both sides are canonicalized, so `..` and symlink spellings are refused too.

**CLI**. `wenlan sources add --type okf <path>`, with the Space header the way `wenlan export okf` sends it, and the add output now prints the queued and waiting counts.

## Verification

- `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo check -p wenlan-app`, `pnpm tsc --noEmit` and `tsc -p tsconfig.tools.json`: clean.
- 34 parser tests over a trimmed copy of `langchain-ai/openwiki` at db1c0f06 (MIT, with the license and commit recorded beside the fixture), plus an overlay for the cases that repository does not contain.
- 5 storage tests, 3 worker tests, 9 sync tests, 3 registration tests, 1 route contract test.
- A mutation proof script: 11 mutations, each killed by the test that claims the behavior.
- Live acceptance against a scratch daemon on its own port and data dir, which aborts unless the daemon reports the scratch pages folder: 31 of 31 checks, including the frontmatter that page detail returns compared against the file's.
- The same run then adds a 1,200-concept bundle, 35 of 35 checks in total. The first sync hands over 1,000 and reports 200 waiting; once the batch is prepared the next sync hands over the rest and the source settles at `queued_files: 0, waiting_files: 0, file_count: 1200, last_sync_errors: 0`. The document lane had to be driven by hand with `POST /api/ambient/sweep`, because the ambient lane is idle-gated and the import priority lane is bounded to 64 slices. That gate is pre-existing and applies to a `directory` source of the same size.

## Review

A read-only Muse review of this diff found that the page-detail `okf` object the plan called for had never been built, which made the provenance write-only. Confirmed and fixed here, with a route test, a mutation, and a live check that compares the returned frontmatter to the file's. Four smaller defects from the same pass are fixed too: an image embed minted a concept link, `status: [deprecated]` imported the concept anyway, a re-poll inflated the source's memory count, and the summary differed with and without a provider. Items deliberately left as follow-ups (a hung-worker lease, `stale_reason` on a resurrected concept, exotic link spellings) are recorded in the plan beside the spec.

## Known limitation

A source page's Space cannot be changed through the daemon's HTTP API today, so the "a page the user moved keeps its Space" behavior is proven by unit tests rather than live.


7xuanlu:
## Scale run: 1,200 concepts

The live acceptance finished at 35 of 35 checks against a scratch daemon on its own port and data dir (`127.0.0.1:7913`, its own `WENLAN_DATA_DIR` and an explicit `knowledge_path`, which the run verifies before creating a page).

| step | result |
|---|---|
| first sync | `files_found: 1200, ingested: 1000, skipped: 0, errors: 0, queued_files: 1000, waiting_files: 200` |
| after the batch is prepared | `files_found: 1200, ingested: 0, skipped: 1200, errors: 0, queued_files: 0, waiting_files: 0` |
| source row | `memory_count: 1200, file_count: 1200, last_sync_errors: 0` |

So the batch gate holds the second batch until the first is prepared, hands the rest over on the next sync, and the counts settle at zero without dropping a file.

### The drain does not happen on its own while you are using the machine

Worth recording because it looked like a hang for 40 minutes. The document lane is idle-gated: `FOREGROUND_INPUT_IDLE_THRESHOLD = 60s` with `idle_samples_required: 2` (`crates/wenlan-server/src/scheduler.rs:44`, `:104`), and `/api/ambient/status` reported `idle_gate_blocked_reason: "ForegroundActive"` the whole time. The import priority lane that skips that gate is bounded by a 10 minute window and `IMPORT_PRIORITY_MAX_AMBIENT_SLICES = 64` shared with inference (`crates/wenlan-server/src/scheduler/import_priority.rs:19-21`), which is exactly why the count stopped at 67 indexed files. Its own doc comment calls that bound intentional: the rest wait for the ordinary idle-gated lap.

The drain above was therefore driven by hand, one document per `POST /api/ambient/sweep`. None of this is new in this PR: a `directory` source of the same size behaves identically, and no code here touches the gate or the lane budget. Flagging it as the honest limit of what the scale check proves, not as a regression.

PR #770, opened 2026-09-20
fix(export): stop three more paths overwriting what a person wrote in their vault

7xuanlu:
Astra's closure review of the integrated OKF arc (#762, #763, #766, #768, #769) found three places the arc did not reach. All three are the mistake the arc exists to remove: evidence that Wenlan **wrote** a path read as proof the bytes there are **still** Wenlan's.

This matters more than an ordinary bug because the projection root is very often the user's live Obsidian vault, so each one is silent data loss in a directory they edit by hand.

## What was wrong

**A cited source stub was truncated on every write.** `project_stubs_for_page_in` opened each stub with `truncate(true)` unconditionally. The orphan GC already archived an *edited* stub, but only once it went orphan. A stub that is still cited, which is the common case, lost any typed-in note on the next write of any page citing that memory.

**`okf_version` on its own was read as a Wenlan signature for `index.md`.** That is the standard OKF root-index shape, not a Wenlan mark. So a hand-written conforming root index from another producer was claimed and overwritten, and so was prose typed into Wenlan's own generated index.

**An OKF stale entry the sweep could not remove was forgotten.** When a directory or symlink stood at a stale path, the sweep skipped it (correctly) and then dropped it from the new marker. The next run had no record the file existed, never retried it, and the bundle kept an obsolete concept for good, even after the obstruction cleared.

## What changed

The stub projection now compares the bytes on disk with the bytes it would write: identical is ours, different is archived first with the same `archive_file_from` the GC calls, and anything that is not a plain file is left alone. No marker can answer this question, because a stub's frontmatter is identical whether or not somebody edited the body. `project_stubs_for_page` now delegates to the same function instead of keeping a second, unprotected copy of the logic.

Index ownership now also requires a body of nothing but the headings and entry lines the renderer emits. The residual is deliberate and recorded at the function: an edit that still looks machine-generated (a reordered or deleted entry line) is still replaced. Closing that needs a digest of the last written index in `state.json`, which is a schema change plus a second state save on every page write, and `regenerate_index_cap` takes `&state` at four call sites.

Unsweepable stale entries are collected, carried into the new marker with their digests, and counted in the existing `ExportStats.failed`. No new wire field.

## Evidence

- `cargo test -p wenlan-core --lib`: **4510 passed, 0 failed**, 33 ignored.
- `cargo clippy -p wenlan-core --lib --all-targets -- -D warnings`: clean.
- `cargo fmt --check`: clean.
- Six new or rewritten tests, each mutation-proved. Reverting a fix's mechanism fails its test:
  - stub classification made inert -> `an_edited_cited_stub_is_archived_before_it_is_rewritten` fails
  - body-shape check made inert -> `a_hand_edit_to_wenlans_own_index_is_not_overwritten` and `a_hand_written_conforming_okf_index_is_left_alone` both fail
  - `unswept` dropped from the marker -> `an_obstructed_stale_entry_stays_in_the_marker_so_a_later_run_retries_it` fails
- `provenance.rs` added to both the `macos` and `windows` `detect-changes` filters, because its new symlink test is `cfg(unix)`. Proved necessary: removing the two lines makes `drift_guard` fail with `unix-sensitive path is not routed through macos`.

## Note on one inverted test

`wenlans_own_index_is_regenerated_after_a_body_edit` asserted the index defect as intended behavior. It is replaced by its inverse, `a_hand_edit_to_wenlans_own_index_is_not_overwritten`, with the reasoning kept in the test's doc comment so a future reader does not restore it. `an_untouched_generated_index_is_still_regenerated` covers the ordinary case that the index still updates.

Spec: `~/.wenlan/specs/wenlan/2026-09-20-okf-closure-review-fixes.md`
Review: `~/.wenlan/specs/wenlan/2026-09-20-okf-arc-astra-closure-review.md`, findings 1, 2 and 5. The remaining six findings are filed separately.


7xuanlu:
## Live-surface evidence

Driven through a real `wenlan-server` on an isolated port (17878), an isolated `WENLAN_DATA_DIR`, and an explicit scratch `knowledge_path` read back from `/api/knowledge/path` before anything was ingested. The user's daemon on :7878 was never touched.

The same script ran against both builds, so the only variable is the code.

| | main `9ccb3427` | this branch `96c16e0b` |
|---|---|---|
| Typed a note into a still-cited stub, then rewrote the page | **DESTROYED** | **PRESERVED** in `archive/mem_a6268c321eac.md` |
| Typed a line into `index.md`, then wrote a second page | **DESTROYED** | **PRESERVED** |
| The second page still projected | yes | yes |

The last row matters as much as the first: the index becoming untouchable does not stop pages from being written. `cold-brew.md` was projected in both runs.

### F1 in detail

Five memories, one page citing all five. Appended `MY NOTE: 94 works better with a light roast than 92.` to `_sources/mem_d9329866d031.md`, then rewrote the page through `PUT /api/pages/{id}`. Afterwards the note is in `archive/mem_d9329866d031.md` byte for byte, and the stub itself is bit-identical to the pristine generated bytes (`cmp` reports identical). On main, `grep -r 'MY NOTE'` over the whole vault returns nothing.

### F2 in detail

The generated index was

```
---
okf_version: "0.2"
---

## Unfiled

* [Pour Over Coffee](/pour-over-coffee.md) - Water temperature, grind, ratio and timing for a pour over.
```

Appended `My reading queue: try the 1:15 ratio next week.` and created a second page. On this branch `index.md` is byte-identical to what was left there; on main the line is gone.

### F5 in detail

`wenlan export okf` into a scratch bundle, then archived Cold Brew so `pages/cold-brew.md` went stale, then replaced that path with a directory holding a file that is not ours.

```
re-export:  {"exported": 1, "skipped": 0, "failed": 1}
marker:     listed: True   digest kept: True
obstruction contents: unchanged
```

Cleared the obstruction and exported again:

```
{"exported": 1, "skipped": 0, "failed": 0}
still listed: False
```

The entry survived in the marker with its digest, was reported rather than silently swallowed, and was swept on the retry. On main it would have dropped out of the marker at the first re-export and never been retried.

### CI

Green at `96c16e0b`, 23 passing / 0 failing, queried by head SHA. One earlier red was `release-preflight (aarch64-apple-darwin)` failing in `ort-sys`'s build script with `failed to lookup address information` for `cdn.pyke.io`, a DNS failure on the runner rather than a test. Per `docs/ci-flake-policy.md` that took exactly one diagnostic rerun, which passed.

7xuanlu:
## Independent review (Muse, via acpx) and what changed

Muse reviewed the integrated diff adversarially, with the live before/after evidence above in hand. It found three real holes and two weak tests. All five are fixed in `3c3a68d2`.

### Real holes

**1. A to-do list was still overwritten.** `* [` is how a Markdown checklist line starts as well as how a rendered index entry starts, so `* [ ] buy milk` under `okf_version`-only frontmatter passed the first body-shape test and the list was replaced by the generated index. An entry line now has to be shaped like one: `* [title](/file.md)`, optionally ` - description`.

**2. A file of bare headings was overwritten.** Every index this projection generates that has a heading also has at least one entry under it, so a file of headings with nothing under them is somebody's outline. It is now left alone.

**3. A source stub this pass cannot read was being archived.** Archiving is a rename, which needs no read permission, so classifying an unreadable file as `Edited` let the pass take over a name whose bytes it had never seen. It is now `Foreign`: left exactly as it is, which is already what the orphan collector does for such a file.

### Test changes

- `a_hand_written_okf_index_with_prose_is_left_alone` was named as though it proved OKF conformance; it checks prose, so it is named for that.
- The older obstruction test now asserts `stats.failed == 1`, not only that the export survived.

### Not changed

Muse also flagged that an index file starting with a UTF-8 BOM fails the frontmatter check. That is pre-existing and not a regression from this branch: `extract_frontmatter` already required `---` at byte 0, so a BOM'd index failed the old check too. Filed separately rather than widened into this PR.

### Checks at `3c3a68d2`

- `cargo fmt -p wenlan-core -- --check`: exit 0
- `cargo clippy -p wenlan-core --lib --all-targets -- -D warnings`: exit 0
- `cargo test -p wenlan-core --lib export::`: 180 passed, 0 failed
- Mutation round over all six mechanisms (stub classification, entry-line shape, headings-only guard, hand-edit detection, marker union, unreadable-stub `Foreign`): every mutation killed

One note on the earlier receipt in this PR: a full-suite run reported one failure in `a_headings_only_outline_is_not_mistaken_for_an_index`. That run compiled while the headings-only mutation was still applied, which fails exactly that one test and nothing else. On clean source the test passes repeatedly, and the full suite is re-running against the pushed commit.

7xuanlu:
## Final receipts at `3c3a68d2`

### CI: green, 23 checks pass

All check runs carry head SHA `3c3a68d22acf700ade071265991743afedc6ebdd`. Run: https://github.com/7xuanlu/wenlan/actions/runs/35526633866

### Full suite

`cargo test -p wenlan-core --lib`: **4513 passed, 0 failed**, 33 ignored (exit 0).

This supersedes the earlier one-failure receipt. That run compiled while the headings-only mutation was still applied; the mutation fails exactly that one test and nothing else, because the sibling checklist test returns early and never reaches the mutated line.

### Live drive, re-run after the Muse changes

The same script against a real daemon on an isolated port, an isolated `WENLAN_DATA_DIR`, and a `knowledge_path` read back from `/api/knowledge/path` before anything was asserted. Five coherent source memories, then a page, then each hostile file typed into the vault by hand, then another page to force a re-projection.

| Check | main `9ccb3427` | fix `3c3a68d2` |
|---|---|---|
| Hand edit in a cited source stub | DESTROYED | **PRESERVED** (moved to `archive/mem_33e98a1a8bb4.md`) |
| Hand-written prose at `index.md` | DESTROYED | **PRESERVED** |
| Checklist at `index.md` (`* [ ] buy milk`) | DESTROYED | **PRESERVED** |
| Headings-only outline at `index.md` | DESTROYED | **PRESERVED** |
| A generated index still regenerates | OK | **OK** |

The last row is the flip side that matters: after removing `index.md`, the next page write regenerates it with all six pages listed, so the index keeps working in the ordinary case.

The user's own daemon was never contacted. Every run used scratch ports 17951 to 17953 and temporary data roots, all removed afterwards.

7xuanlu:
## Closure review (Fable, lead) of the integrated diff at `3c3a68d2`

Read all four files against main `9ccb3427`. Three findings, all fixed in `888bf3d9`; one residual recorded, not blocking.

### Fixed

**1. Two fail-open stat errors.** `stub_target` (provenance.rs) turned *any* `symlink_metadata` error into `Writable`, and the OKF stale sweep (okf.rs) turned *any* error into "already gone". Only `NotFound` means that. A directory this process cannot search, or an I/O error, says nothing about what is at the name, and both callers then act as if it were free: the stub writer opens with `create(true).truncate(true)`, and the sweep drops the entry from the marker, which is the exact forgetting the sweep change exists to stop. Both now fail closed: the stub is left alone (`Foreign`), and the entry stays listed with `failed += 1`. No test provokes a non-`NotFound` stat error portably, so this edge is a strict narrowing of when the pass writes, not a proven behavior. Said plainly so nobody reads it as covered.

**2. Doc splice.** The new shape test was inserted between `index_entry_line` and its doc comment, so rustdoc stacked three comment blocks (one duplicated) on `index_body_is_generated` and left `index_entry_line` undocumented. Restored.

**3. CLI exit message.** `stats.failed` now also counts stale paths the sweep could not remove, and `wenlan export okf` exits non-zero on `failed > 0` with "failed page(s)". That wording was wrong for a sweep obstruction; it now says "path(s) it could not write or sweep". The non-zero exit itself is a deliberate behavior change: a bundle that still carries an obsolete concept is not the bundle that was asked for.

### Residual, recorded

`index_line_is_entry` splits on the first `](` and the first `)`. Projection filenames are alnum plus `-` (`slugify`), so the filename side is safe; a page *title* containing `](` followed by `)` would make its own rendered line unparseable, the index would be treated as not ours, and it would stop updating with the existing warning. Safe direction, no data loss, very rare title shape. Not changed.

### Checked

- `cargo fmt -p wenlan-core -p wenlan -- --check`: exit 0
- `cargo clippy -p wenlan-core -p wenlan --all-targets -- -D warnings`: exit 0
- `cargo test -p wenlan-core --lib export::`: 180 passed, 0 failed
- Everything else in the earlier receipts (full suite 4513/0, the five-row live before/after) was run at `3c3a68d2`; the delta to `888bf3d9` is the three items above. CI at `888bf3d9` is the gate for the merge.

